PR #26, opened 2022-07-29
feat: impl ceresmeta grpc service & redirect request to leader

### LINKED ISSUE #18: Redirect requests to leader
**Description**
The cluster of ceresmeta provides service as a whole for ceresdb nodes, that is to say, one request can be received by any ceresmeta instance.
However, some service, e.g. ID generator, actually can only be provided by the leader. So the followers must can redirect the traffic into the leaders for requests of specific types.

**Proposal**
(TODO)

**Additional context**

chunshao90:

Closes #18 

# Rationale for this change
 
The cluster of ceresmeta provides service as a whole for ceresdb nodes, that is to say, one request can be received by any ceresmeta instance.
However, some service, e.g. ID generator, actually can only be provided by the leader. So the followers must can redirect the traffic into the leaders for requests of specific types.

* impl ceresmeta GRPC service, defined in https://github.com/CeresDB/ceresdbproto/blob/12a0ee8e45883056f275c707518fa9ed508676d1/protos/meta_service.proto#L10
* forward request to the leader

none

# How does this change test
TODO:  need to add tests


server/grpcservice/util.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/CeresDB/ceresdbproto/pkg/commonpb"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// GetClientConn returns a gRPC client connection.
+func GetClientConn(ctx context.Context, addr string, tlsCfg *tls.Config, do ...grpc.DialOption) (*grpc.ClientConn, error) {
+	var creds credentials.TransportCredentials
```
ShiKaiWi, line 18: Now the tls is not necessary.

server/server.go
```diff
@@ -192,6 +203,6 @@ func (srv *Server) UnbindHeartbeatStream(_ context.Context, node string) error {
 	return nil
 }
 
-func (*Server) ProcessHeartbeat(_ context.Context, _ *metaservicepb.NodeHeartbeatRequest) error {
-	return nil
+func (srv *Server) ProcessHeartbeat(ctx context.Context, req *metaservicepb.NodeHeartbeatRequest) error {
+	return srv.clusterManager.RegisterNode(ctx, req.GetHeader().GetClusterName(), req.GetInfo().GetNode(), req.GetInfo().Lease)
```
ShiKaiWi, line 207: Why not get `Lease` by `GetLease`?

server/server.go
```diff
@@ -134,6 +137,10 @@ func (srv *Server) startEtcd(ctx context.Context) error {
 /// startServer starts involved services.
 func (srv *Server) startServer(ctx context.Context) error {
 	srv.hbStreams = schedule.NewHeartbeatStreams(ctx)
+	storage := storage.NewStorageWithEtcdBackend(srv.etcdCli, srv.cfg.StorageRootPath, storage.Options{
+		MaxScanLimit: 100, MinScanLimit: 10,
```
ShiKaiWi, line 141: Avoid the hard code.

server/grpcservice/service.go
```diff
@@ -5,24 +5,32 @@ package grpcservice
...
 
 type Service struct {
 	metaservicepb.UnimplementedCeresmetaRpcServiceServer
-
 	opTimeout time.Duration
 	h         Handler
+	manager   cluster.Manager
+
+	// Store as map[string]*grpc.ClientConn
+	grpcClientConns sync.Map
```
ShiKaiWi, line 26: - How about `connCache`?
- It seems there is no way to clear invalid connections and maybe we should take considerations for that.

server/grpcservice/service.go
```diff
@@ -5,24 +5,32 @@ package grpcservice
...
+	"github.com/pkg/errors"
 	"go.uber.org/zap"
 )
 
 type Service struct {
 	metaservicepb.UnimplementedCeresmetaRpcServiceServer
-
 	opTimeout time.Duration
 	h         Handler
+	manager   cluster.Manager
```
ShiKaiWi, line 23: The service seems should not hold the manager which should be held by `server`(`h` in this service).

server/grpcservice/service.go
```diff
@@ -105,6 +131,50 @@ func (s *Service) NodeHeartbeat(heartbeatSrv metaservicepb.CeresmetaRpcService_N
 			return ErrRecvHeartbeat.WithCause(err)
 		}
 
+		// Forward request to the leader.
+		{
```
ShiKaiWi, line 135: Guess we can define this block as a new method.

server/grpcservice/util.go
chunshao90: grpc.WithInsecure() is  deprecated. In addition, this is a util func, so i expose tlsCfg.

pkg/coderr/error.go
```diff
@@ -23,16 +23,25 @@ type CodeError interface {
...
+		return true
+	}
+
+	return false
+}
+
+func GetCauseCode(err error) (Code, bool) {
 	if err == nil {
-		return false
+		return Ok, false
```
ShiKaiWi, line 36: Just return 0 and no need to introduce a OK code.

server/cluster/cluster.go
```diff
@@ -210,20 +219,20 @@ func (c *Cluster) CreateSchema(ctx context.Context, schemaName string, schemaID
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	// check if exists
+	// Check provided schema if exists.
```
ShiKaiWi, line 222:
```suggestion
	// Check if provided schema exists.
```

server/grpcservice/forward.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if forwardedAddr != "" {
+		ceresmetaClient, err := s.getCeresmetaClient(ctx, forwardedAddr)
+		if err != nil {
+			return nil, errors.Wrap(err, "get forwarded ceresmeta client")
+		}
+		return ceresmetaClient, nil
+	}
+	return nil, nil
```
ShiKaiWi, line 26: Maybe we should add comments to explain this case.

server/grpcservice/util.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"net/url"
+
+	"github.com/CeresDB/ceresdbproto/pkg/commonpb"
+	"github.com/CeresDB/ceresmeta/pkg/coderr"
+	"google.golang.org/grpc"
+)
+
+// GetClientConn returns a gRPC client connection.
+func GetClientConn(ctx context.Context, addr string) (*grpc.ClientConn, error) {
```
ShiKaiWi, line 15: No need to expose these methods (GetClientConn, OkResponseHeader, ResponseHeader).

pkg/coderr/error.go
```diff
@@ -23,16 +23,25 @@ type CodeError interface {
...
 	if err == nil {
-		return false
+		return 0, false
 	}
 
 	cause := errors.Cause(err)
 	cerr, ok := cause.(CodeError)
 	if !ok {
-		return false
+		return Internal, false
```
ShiKaiWi, line 42:
```suggestion
		return 0, false
```

ShiKaiWi (approved):

PR #41, opened 2022-08-24
feat: support leadership event callback mechanism

ShiKaiWi:

Closes #

# Rationale for this change
 Some data is exclusive and can only be kept by the leader, that is to say, these data can't be accessed by followers and leaders at the same time. Such a requirement is a little difficult to be met when the leadership is being transferred.

What we need is actually is a mechanism to enforce then order of the leadership transference between the old leader and the new leader so that we can disallow the old leader access to the exclusive data before the old leader transfers the leadership and allow the new leader access to the data before the new leader takes the leadership.

 

Introduce the `LeadershipEventCallbacks` to the `Leadership.Watch` method and the callbacks will be called before the leader hands over the leadership and after the leader takes up the leadership.

None.

# How does this change test
There is no test yet, but it can be tested actually. Some tests should be supplemented afterwards.

### DIFF
--- a/server/member/lease.go
+++ b/server/member/lease.go
@@ -38,6 +38,22 @@ func newLease(rawLease clientv3.Lease, ttlSec int64) *lease {
 	}
 }
 
+type renewLeaseResult int
+
+const (
+	renewLeaseSuccessful renewLeaseResult = iota
+	renewLeaseFailed
+	renewLeaseExpired
+)
+
+func (r renewLeaseResult) failed() bool {
+	return r == renewLeaseFailed
+}
+
+func (r renewLeaseResult) successful() bool {
+	return r == renewLeaseSuccessful
+}
+
 func (l *lease) Grant(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctx, l.timeout)
 	defer cancel()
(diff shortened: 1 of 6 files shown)


server/member/watch_leader.go
```diff
@@ -47,7 +52,9 @@ func NewLeaderWatcher(ctx WatchContext, self *Member, leaseTTLSec int64) *Leader
 //   - Elect the etcd leader as the ceresmeta leader.
 //	 - The leader keeps the leadership lease alive.
 //   - The other members keeps waiting for the leader changes.
-func (l *LeaderWatcher) Watch(ctx context.Context) {
+//
+// The LeadershipCallbacks callback will be triggered when specific events occur.
```
chunshao90, line 56: `The LeadershipCallbacks will be triggered when specific events occur.`

server/member/lease.go
```diff
@@ -142,41 +161,47 @@ func (l *lease) getExpireTime() time.Time {
...
 		// init the timer for next keep alive action.
 		t := time.After(interval)
 
-		// notify success of the renewed event.
-		select {
-		case renewed <- struct{}{}:
-		case <-ctx.Done():
-			break L
+		if !renewRes.failed() {
+			// notify result of the renew.
```
chunshao90, line 197: `Notify`

server/member/lease.go
```diff
@@ -142,41 +161,47 @@ func (l *lease) getExpireTime() time.Time {
...
 			expireAt := start.Add(time.Duration(resp.TTL) * time.Second)
 			updated := l.setExpireTimeIfNewer(expireAt)
 			l.logger.Debug("got next expired time", zap.Time("expired-at", expireAt), zap.Bool("updated", updated))
-		}()
+			return renewLeaseSuccessful
+		}
+
+		renewRes := renewOnce()
 
 		// init the timer for next keep alive action.
```
chunshao90, line 193: `Init`

chunshao90 (approved):
LGTM

PR #55, opened 2022-09-13
refactor: attach call stack to coderr

### LINKED ISSUE #20: attach backtrace (callstack) to the leaf errors
**Description**
Now the leaf error generated by coderr pkg has no inforamtion about the backtrace (callstack) which is important for trobleshooting problems.

**Proposal**
Generate the call stacks and attach it to the leaf error.
We may refer to [implementations of errors pkg](https://github.com/pkg/errors/blob/master/stack.go#L163).

**Additional context**

ShiKaiWi:

Closes #20 

# Rationale for this change
 The call stack should be attached to the error cause, which helps a lot when troubleshooting problems.
 

- Replace `errors.Wrap(f)` with `errors.WithMessage(f)`;
- Attach call stack to `coderr` cause;
- Some fixes for docs;

None.

# How does this change test
New ut.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,14 +5,14 @@ Closes #
 # Rationale for this change
  
  <!---
- Why are you proposing this change? If this is already explained clearly in the issue then this section is not needed.
+ Why are you proposing this change? If this is already explained clearly in the issue, then this section is not needed.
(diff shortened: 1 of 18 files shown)


chunshao90 (approved):
LGTM

PR #58, opened 2022-09-21
feat: define procedure interface

chunshao90:

Closes #

# Rationale for this change
 `Procedure` is used to describe how to execute a set of operations of the procedure.
 All procedures like `SwitchLeaderProcedure` and `MergeShardProcedure` need to implement `Procedure`.

 

Define `Procedure` interface.

No.

# How does this change test
No.

### DIFF
--- a/server/coordinator/procedure/procedure.go
+++ b/server/coordinator/procedure/procedure.go
@@ -1,0 +1,45 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+package procedure
+
+import "context"
+
+type State string
+
+const (
+	StateRunning   = "running"
+	StateFinished  = "finished"
+	StateFailed    = "failed"
+	StateCancelled = "cancelled"
+)
+
+type procedureType = uint
+
+const (
+	switchLeader procedureType = iota
+	mergeShard
+)
+
+// Procedure is used to describe how to execute a set of operations from the scheduler, e.g. SwitchLeaderProcedure, MergeShardProcedure.
+type Procedure interface {
+	// ID of the procedure.
+	ID() uint64
+
+	// Type of the procedure.
+	Type() procedureType
+
+	// Start the procedure.
+	Start(ctx context.Context) error
+
+	// Cancel the procedure.
+	Cancel(ctx context.Context) error
+
+	// State of the procedure. Retrieve the state of this procedure.
+	State() State
+}
+
+// nolint
+type Manager struct {
+	storage    *Storage
+	procedures []Procedure
+}
--- a/server/coordinator/procedure/procedure.go
+++ b/server/coordinator/procedure/procedure.go
@@ -13,10 +13,10 @@ const (
 	StateCancelled = "cancelled"
 )
 
-type procedureType = uint
+type Typ = uint
 
 const (
-	switchLeader procedureType = iota
+	switchLeader Typ = iota
 	mergeShard
 )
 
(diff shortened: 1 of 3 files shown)


server/coordinator/procedure/storage.go
```diff
@@ -0,0 +1,21 @@
+package procedure
+
+import (
+	"context"
+)
+
+type Write interface {
+	CreateOrUpdate(ctx context.Context, meta *Meta) error
```
ZuLiangWang, line 8: Because the `state` parameter is used in  the `list` func, it should also be used as an independent parameter here, or the `state` parameter in the `list` func should be removed directly

server/coordinator/procedure/switch_leader.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"context"
+)
+
+func NewSwitchLeaderProcedure() *SwitchLeaderProcedure {
+	return nil
+}
+
+func LoadSwitchLeaderProcedure(_ []byte) *SwitchLeaderProcedure {
```
ZuLiangWang: Procedure should not process `[]byte` directly, `List` func in `Storage` Process `[]byte` and return `*Meta`, so the parameter here should be `*Meta`

server/coordinator/procedure/procedure.go
```diff
@@ -0,0 +1,45 @@
...
+	StateFinished  = "finished"
+	StateFailed    = "failed"
+	StateCancelled = "cancelled"
+)
+
+type procedureType = uint
+
+const (
+	switchLeader procedureType = iota
+	mergeShard
```
ZuLiangWang, line 20: ShardOperationType_Create         ShardOperationType = 0
	ShardOperationType_Delete         ShardOperationType = 1
	ShardOperationType_TransferLeader ShardOperationType = 2
	ShardOperationType_Migrate        ShardOperationType = 3
	ShardOperationType_Split          ShardOperationType = 4
	ShardOperationType_Merge          ShardOperationType = 5

server/coordinator/procedure/procedure.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// Procedure is used to describe how to execute a set of operations from the scheduler, e.g. SwitchLeaderProcedure, MergeShardProcedure.
+type Procedure interface {
+	// ID of the procedure.
+	ID() uint64
+
+	// Type of the procedure.
+	Type() procedureType
```
ShiKaiWi, line 29: As the return value of an exported interface should be public too, so it should be `ProcedureType`.

server/coordinator/procedure/storage.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// nolint
+type Meta struct {
```
ShiKaiWi, line 12: The fields should be public, otherwise some encoding function may not work.

server/coordinator/procedure/storage.go
```diff
@@ -0,0 +1,22 @@
...
+	id      uint64
+	typ     procedureType
+	state   State
+	rawData []byte
+}
+
+type Storage interface {
+	Write
+	List(ctx context.Context, state State) (*Meta, error)
```
ShiKaiWi, line 21: - The `state` parameter maybe prevents callers to fetch all the Procedure Meta from the storage.
- Should the returned value be `[]*Meta`?

BTW, maybe we should revise the method signature to something like https://github.com/CeresDB/ceresmeta/blob/5eb015274de053acc027ce5cec5f22de795b60af/server/etcdutil/util.go#L27

server/coordinator/procedure/storage.go
```diff
@@ -0,0 +1,22 @@
+package procedure
```
ShiKaiWi, line 1: License header seems missing.

ShiKaiWi (approved):

Issue #62, opened 2022-09-27
Add data persistence implementation for procedure


ZuLiangWang:
**Description**
In order to achieve disaster recovery of shard operation, we need to support the persistent storage of `Procedure`. When CeresMeta restarts each time, the data of `Procedure` is loaded and scheduled.

**Proposal**
- [ ] Implementation of procedure storage based on `etcd`, and expose appropriate interfaces externally.
- [ ] Use the interface provided by storage in Procedure to realize data persistence.
- [ ] Reload and continue to execute Procedure when the CeresMeta node restarts.

**Additional context**

PR #68, opened 2022-09-30
feat: add procedure storage implementation

ZuLiangWang:
https://github.com/CeresDB/ceresmeta/issues/62

# Rationale for this change
Implementation of procedure storage based on `etcd`, and expose appropriate interfaces externally.

* Add `Delete` func to `Write` interface
* Add implementation for `EtcdStorageImpl`
* Add test for `EtcdStorageImpl`

None.

# How does this change test
Pass unit test.

### DIFF
--- a/server/coordinator/procedure/storage_impl.go
+++ b/server/coordinator/procedure/storage_impl.go
@@ -13,40 +13,60 @@ import (
 )
 
 const (
-	Version       = "v1"
-	PathProcedure = "procedure"
+	Version              = "v1"
+	PathProcedure        = "procedure"
+	PathHistoryProcedure = "historyProcedure"
 )
 
 type EtcdStorageImpl struct {
-	client *clientv3.Client
-
-	// opts Options
+	client    *clientv3.Client
 	clusterId uint32
-
-	rootPath string
+	rootPath  string
 }
 
-// example:
-// procedure 1: v1/procedure/{ClusterID}/{procedureID} -> {procedureType} + {procedureState} + {data}
+func NewEtcdStorageImpl(client *clientv3.Client, clusterId uint32, rootPath string) *EtcdStorageImpl {
+	return &EtcdStorageImpl{
+		client:    client,
+		clusterId: clusterId,
+		rootPath:  rootPath,
+	}
+}
 
+// CreateOrUpdate example:
+// procedure : /{rootPath}/v1/procedure/{clusterID}/{procedureID} -> {procedureType} + {procedureState} + {data}
 func (e EtcdStorageImpl) CreateOrUpdate(ctx context.Context, meta *Meta) error {
-	keyPath := e.generateKeyPath(meta.ID)
-
 	str, err := encode(meta)
 	if err != nil {
 		return errors.WithMessage(err, "encode meta failed")
 	}
-	opPutProcedure := clientv3.OpPut(keyPath, str)
+	keyPath := e.generateKeyPath(meta.ID, false)
+	opPut := clientv3.OpPut(keyPath, str)
 
 	_, err = e.client.Txn(ctx).
-		Then(opPutProcedure).
+		Then(opPut).
 		Commit()
 
 	return err
 }
 
-func (e EtcdStorageImpl) Scan(ctx context.Context, batchSize uint, state State, typ Typ) ([]*Meta, error) {
+// Delete example:
+// historyProcedure : /{rootPath}/v1/historyProcedure/{clusterID}/{procedureID}
+func (e EtcdStorageImpl) Delete(ctx context.Context, meta *Meta) error {
+	str, err := encode(meta)
+	if err != nil {
+		return errors.WithMessage(err, "encode meta failed")
+	}
+	keyPath := e.generateKeyPath(meta.ID, false)
+	historyKeyPath := e.generateKeyPath(meta.ID, true)
+	opDelete := clientv3.OpDelete(keyPath)
+	opPut := clientv3.OpPut(historyKeyPath, str)
 
+	_, err = e.client.Txn(ctx).Then(opDelete).Then(opPut).Commit()
+
+	return err
+}
+
+func (e EtcdStorageImpl) Scan(ctx context.Context, batchSize uint, state State, typ Typ) ([]*Meta, error) {
 	metas := make([]*Meta, 0)
 	do := func(_ string, value []byte) error {
 		meta := &Meta{}
@@ -58,8 +78,8 @@ func (e EtcdStorageImpl) Scan(ctx context.Context, batchSize uint, state State, 
 		return nil
 	}
 
-	startKey := e.generateKeyPath(0)
-	endKey := e.generateKeyPath(math.MaxUint64)
+	startKey := e.generateKeyPath(0, false)
+	endKey := e.generateKeyPath(math.MaxUint64, false)
 
 	err := etcdutil.Scan(ctx, e.client, startKey, endKey, math.MaxInt, do)
 	if err != nil {
@@ -68,11 +88,14 @@ func (e EtcdStorageImpl) Scan(ctx context.Context, batchSize uint, state State, 
 	return metas, nil
 }
 
-/*
-*
- */
-func (e EtcdStorageImpl) generateKeyPath(procedureId uint64) string {
-	return path.Join(e.rootPath, Version, PathProcedure, fmtID(uint64(e.clusterId)), fmtID(procedureId))
+func (e EtcdStorageImpl) generateKeyPath(procedureId uint64, isHistory bool) string {
+	var pathPrefix string
+	if isHistory {
+		pathPrefix = PathHistoryProcedure
+	} else {
+		pathPrefix = PathProcedure
+	}
+	return path.Join(e.rootPath, Version, pathPrefix, fmtID(uint64(e.clusterId)), fmtID(procedureId))
 }
 
 func fmtID(id uint64) string {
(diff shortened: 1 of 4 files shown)


server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return path.Join(e.rootPath, Version, pathPrefix, fmtID(uint64(e.clusterID)), fmtID(procedureID))
+}
+
+func fmtID(id uint64) string {
+	return fmt.Sprintf("%020d", id)
+}
+
+func encode(meta *Meta) (string, error) {
+	bytes, err := json.Marshal(meta)
```
ShiKaiWi, line 108: `Meta` contains a field whose type is ray byte, and that is to say, it may be not suitable to encode it in json format.

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,118 @@
...
+	historyKeyPath := e.generateKeyPath(meta.ID, true)
+	opDelete := clientv3.OpDelete(keyPath)
+	opPut := clientv3.OpPut(historyKeyPath, str)
+
+	_, err = e.client.Txn(ctx).Then(opDelete, opPut).Commit()
+
+	return err
+}
+
+func (e EtcdStorageImpl) Scan(ctx context.Context, batchSize int) ([]*Meta, error) {
```
ShiKaiWi, line 71: It seems `ReadAll` is a better name for this method.

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,118 @@
...
+	str, err := encode(meta)
+	if err != nil {
+		return errors.WithMessage(err, "encode meta failed")
+	}
+	keyPath := e.generateKeyPath(meta.ID, false)
+	opPut := clientv3.OpPut(keyPath, str)
+
+	_, err = e.client.Txn(ctx).
+		Then(opPut).
+		Commit()
```
ShiKaiWi, line 49: Is it necessary to use txn?

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/CeresDB/ceresmeta/server/etcdutil"
+	"github.com/pkg/errors"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+const (
+	Version              = "v1"
+	PathProcedure        = "procedure"
+	PathHistoryProcedure = "historyProcedure"
```
ShiKaiWi, line 20:
```suggestion
	PathDeletedProcedure = "deletedProcedure"
```

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,118 @@
...
+		if err := decode(meta, string(value)); err != nil {
+			return errors.WithMessage(err, "decode meta failed")
+		}
+
+		metas = append(metas, meta)
+		return nil
+	}
+
+	startKey := e.generateKeyPath(uint64(0), false)
+	endKey := e.generateKeyPath(math.MaxUint64, false)
```
ShiKaiWi, line 84: Add two more methods called to avoid the unreadable boolean:
```
generateNormalKeyPath
generateDeletedKeyPath
```

server/coordinator/procedure/storage_impl.go
ZuLiangWang: `encode` function accepts a `meta` instance as a parameter. The non-utf8 data of the procedure is stored in the `rawdata` in the `meta` as `[]bytes`. Storage is only responsible for reading and writing the meta. How to convert all member variables in the `procedure` to `[]bytes` is not implemented in this PR, but is implemented by each `procedure` itself.

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,122 @@
...
+
+	err := etcdutil.Scan(ctx, e.client, startKey, endKey, batchSize, do)
+	if err != nil {
+		return errors.WithMessage(err, "scan procedure failed")
+	}
+	return nil
+}
+
+func (e EtcdStorageImpl) generateNormalKeyPath(procedureID uint64) string {
+	return path.Join(e.rootPath, Version, PathProcedure, fmtID(uint64(e.clusterID)), fmtID(procedureID))
```
ShiKaiWi, line 100: Why not keep the original method: `generateKeyPath(id uint64, isDeleted bool)`? And then do such call here:
```go
return generateKeyPath(procedureID, false);
```

As for the `generateDeletedKeyPath`, put it as:
```go
return generateKeyPath(procedureID, true);
```

server/coordinator/procedure/storage_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+func testScan(t *testing.T, storage *EtcdStorageImpl) {
+	re := require.New(t)
+	ctx, cancel := context.WithTimeout(context.Background(), DefaultTimeout)
+	defer cancel()
+
+	metas := make([]*Meta, 0)
+	err := storage.ReadAll(ctx, DefaultScanBatchSie, &metas)
+	re.NoError(err)
+	re.Equal(2, len(metas))
```
ShiKaiWi, line 66: Maybe it is not enough to do only length check, and shall do a more content check on the `metas`?

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,122 @@
...
+	opPut := clientv3.OpPut(keyPath, str)
+
+	if _, err = e.client.Do(ctx, opPut); err != nil {
+		return errors.WithMessage(err, "etcd put data failed")
+	}
+	return nil
+}
+
+// Delete example:
+// historyProcedure : /{rootPath}/v1/historyProcedure/{clusterID}/{procedureID}
```
ShiKaiWi, line 62: Put all the comments as a statement as possible as you can:
```suggestion
// Do a soft deletion, and the deleted key's format is:
//  /{rootPath}/v1/historyProcedure/{clusterID}/{procedureID}
```

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,122 @@
...
+func (e EtcdStorageImpl) CreateOrUpdate(ctx context.Context, meta *Meta) error {
+	state := meta.State
+	if state == StateFinished || state == StateCancelled {
+		if err := e.deleteAndRecord(ctx, meta); err != nil {
+			return errors.WithMessage(err, "etcd delete data failed")
+		}
+		return nil
+	}
+
+	str, err := encode(meta)
```
ShiKaiWi, line 48: `str` is weird (it is known that its type is string), and how about just call it as `s` for such a temporal variable?

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (e EtcdStorageImpl) generateDeletedKeyPath(procedureID uint64) string {
+	return path.Join(e.rootPath, Version, PathDeletedProcedure, fmtID(uint64(e.clusterID)), fmtID(procedureID))
+}
+
+func fmtID(id uint64) string {
+	return fmt.Sprintf("%020d", id)
+}
+
+func encode(meta *Meta) (string, error) {
```
ShiKaiWi, line 111: Maybe we should use protobuf to do such encoding for `Meta` for a good compatibility in the future. You can refer to the examples in `server/storage/storage_impl.go`.

ZuLiangWang: Resolve this conversation, left a `TODO` comment.

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,123 @@
...
+	opPut := clientv3.OpPut(keyPath, s)
+
+	if _, err = e.client.Do(ctx, opPut); err != nil {
+		return errors.WithMessage(err, "etcd put data failed")
+	}
+	return nil
+}
+
+// Do a soft deletion, and the deleted key's format is:
+// deletedProcedure : /{rootPath}/v1/historyProcedure/{clusterID}/{procedureID}
```
ShiKaiWi, line 62:
```suggestion
//  /{rootPath}/v1/historyProcedure/{clusterID}/{procedureID}
```

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,123 @@
...
+func NewEtcdStorageImpl(client *clientv3.Client, clusterID uint32, rootPath string) *EtcdStorageImpl {
+	return &EtcdStorageImpl{
+		client:    client,
+		clusterID: clusterID,
+		rootPath:  rootPath,
+	}
+}
+
+// CreateOrUpdate example:
+// procedure : /{rootPath}/v1/procedure/{clusterID}/{procedureID} -> {procedureType} + {procedureState} + {data}
```
ShiKaiWi, line 38: Polish up the comment just like `deleteAndRecord`.

server/coordinator/procedure/storage_impl.go
ZuLiangWang: This is a potential problem, we cloud resolve it by:
1. Constrain the data written in `rawdata` to be in utf8 format.
2. Replace json with proto.
I left TODO here and just resolve this conversation.

server/coordinator/procedure/storage_impl.go
```diff
@@ -0,0 +1,123 @@
...
+func (e EtcdStorageImpl) CreateOrUpdate(ctx context.Context, meta *Meta) error {
+	state := meta.State
+	if state == StateFinished || state == StateCancelled {
```
ShiKaiWi, line 41: Shall we define a method on `State` called `NeedRetry`?

server/coordinator/procedure/storage_impl.go
ShiKaiWi: I'm OK with replace json with proto.

ShiKaiWi (approved):

PR #78, opened 2022-10-19
feat: add DropTableProcedure

chunshao90:

Closes #

# Rationale for this change
Add DropTableProcedure implementation.
 

Add DropTableProcedure implementation.

No.

# How does this change test
Add `TestCreateAndDropTable` test to test it.

### DIFF
--- a/server/coordinator/procedure/drop_table.go
+++ b/server/coordinator/procedure/drop_table.go
@@ -6,6 +6,10 @@ import (
 	"context"
 	"sync"
 
+	"go.uber.org/zap"
+
+	"github.com/CeresDB/ceresmeta/pkg/log"
+
 	"github.com/CeresDB/ceresdbproto/pkg/metaservicepb"
 	"github.com/CeresDB/ceresmeta/server/cluster"
 	"github.com/CeresDB/ceresmeta/server/coordinator/eventdispatch"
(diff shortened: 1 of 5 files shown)


server/coordinator/procedure/drop_table.go
```diff
@@ -0,0 +1,170 @@
...
+
+func dropTablePrepareCallback(event *fsm.Event) {
+	request := event.Args[0].(*DropTableCallbackRequest)
+	table, exists, err := request.cluster.GetTable(request.ctx, request.schemaName, request.tableName)
+	if err != nil {
+		cancelEventWithLog(event, err, "cluster get table")
+		return
+	}
+	if !exists {
+		return
```
ZuLiangWang, line 48:
```suggestion
                 log.warn("drop table stop because table not found",zap.uint64("tableName",request.tableName))
		return
```

chunshao90: If the table does not exist, `DropTableProcedure` runs successfully.
I don't think there is a need to print the log here.

ShiKaiWi: Dropping a non-existent table may be a corner case when running, and logging for that may helps debugging if any problems happen.

ShiKaiWi (approved):
LGTM

PR #87, opened 2022-10-25
fix: update the cache when create shard topology

ShiKaiWi:

Closes #

# Rationale for this change
 For now, cache won't be updated when creating new shard topologies, which leading to multiple `Load` on `Cluster` but such load operation is not allowed to be called more than once.

 

- Update the cache when creating shard topologies;
- Remove multiple Load op;

None.

# How does this change test
Test by manual.


chunshao90 (approved):
LGTM

PR #100, opened 2022-11-16
chore: bump ceresdbproto to the latest version

archerny:

Closes #

# Rationale for this change
 
Dirs of underlying dependency `ceresdbproto` has been changed, bump to the latest version and do some related adaption

1. Dependency description for `ceresdbproto` has been modified.
2. Import path of several packages has been adpapted.

nope.

# How does this change test

Pass the existing CI.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,9 @@ go 1.19
 
 require (
 	github.com/AlekSi/gocov-xml v1.0.0
+	github.com/CeresDB/ceresdbproto/golang v0.0.0-20221115062517-8cd021e19e44
 	github.com/axw/gocov v1.1.0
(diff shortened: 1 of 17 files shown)


go.mod
```diff
@@ -24,6 +23,7 @@ require (
 
 require (
 	github.com/BurntSushi/toml v1.1.0 // indirect
+	github.com/CeresDB/ceresdbproto/golang v0.0.0-20221115062517-8cd021e19e44 // indirect
```
ZuLiangWang, line 26: This dependency should move to previous require same as before, because this dependency is not an indirect dependency.

ZuLiangWang (approved):
LGTM

PR #101, opened 2022-11-17
chore: add toml config parser

ZuLiangWang:
Closes #

# Rationale for this change
At present, ceresmeta's startup parameters only support specifying in program arguments. These configurations are difficult to maintain. We want to specify startup parameters by setting configuration files. 

* Add toml config parser, it will parse toml config when ceresmeta starts.
* Add a sample toml config example.

User can configure the startup parameters by input toml file path now.

# How does this change test
Local manual test.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -51,7 +51,7 @@ const (
 )
 
 type Config struct {
-	Config string `toml:"config" json:"config"`
+	Path string `toml:"path" json:"path"`
 
 	Log     log.Config `toml:"log" json:"log"`
 	EtcdLog log.Config `toml:"etcd-log" json:"etcd-log"`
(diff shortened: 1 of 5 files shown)


server/config/config.go
```diff
@@ -48,6 +51,8 @@ const (
 )
 
 type Config struct {
+	Config string `toml:"config" json:"config"`
```
ShiKaiWi, line 54: Shall we call it `Path`?

server/config/config.go
```diff
@@ -250,3 +257,24 @@ func MakeConfigParser() (*Parser, error) {
 	fs.IntVar(&cfg.HTTPPort, "http-port", defaultHTTPPort, "port of http server")
 	return builder, nil
 }
+
+// MakeConfigParseFromToml read configuration from the toml file, if the config item already exists, it will be overwritten.
+func MakeConfigParseFromToml(conf *Config) error {
```
ShiKaiWi, line 262: We need a better name:
```suggestion
func ParseConfigFromFile(conf *Config) error {
```

BTW, will this cause start failure if the command line doesn't contain the config file path?

ZuLiangWang: > We need a better name:
> 
> BTW, will this cause start failure if the command line doesn't contain the config file path?

Server will skip config parse if config file path not specified, it will be started in the way that there is no configuration file before.

ShiKaiWi (approved):
LGTM

PR #102, opened 2022-11-18
fix: fix http server start

ZuLiangWang:

Closes #

# Rationale for this change
Now the http server will be started when the meta starts synchronously, this will cause the meta server to fail to initialize normally.

* Change the start of http server from synchronous to asynchronous.

None.

# How does this change test
Local manual test.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -184,10 +184,14 @@ func (srv *Server) startServer(_ context.Context) error {
 
 	api := http.NewAPI(procedureManager, procedureFactory, manager)
 	httpService := http.NewHTTPService(srv.cfg.HTTPPort, time.Second*10, time.Second*10, api.NewAPIRouter())
-	err = httpService.Start()
-	if err != nil {
-		return errors.WithMessage(err, "start http service failed")
-	}
+	go func() {
+		err := httpService.Start()
+		if err != nil {
+			if err != nil {
+				log.Error("start http service failed", zap.Error(err))
+			}
+		}
+	}()
 	srv.httpService = httpService
 
 	log.Info("server started")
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -184,10 +184,14 @@ func (srv *Server) startServer(_ context.Context) error {
...
 	api := http.NewAPI(procedureManager, procedureFactory, manager)
 	httpService := http.NewHTTPService(srv.cfg.HTTPPort, time.Second*10, time.Second*10, api.NewAPIRouter())
-	err = httpService.Start()
-	if err != nil {
-		return errors.WithMessage(err, "start http service failed")
-	}
+	go func() {
+		err := httpService.Start()
+		if err != nil {
+			if err != nil {
```
ShiKaiWi, line 190: Remove this redundant check.

server/server.go
```diff
@@ -184,10 +184,14 @@ func (srv *Server) startServer(_ context.Context) error {
 
 	api := http.NewAPI(procedureManager, procedureFactory, manager)
 	httpService := http.NewHTTPService(srv.cfg.HTTPPort, time.Second*10, time.Second*10, api.NewAPIRouter())
-	err = httpService.Start()
-	if err != nil {
-		return errors.WithMessage(err, "start http service failed")
-	}
+	go func() {
+		err := httpService.Start()
```
ShiKaiWi, line 188: Any control method to close the http service?

ZuLiangWang: `http.Stop()` has been definition but not used. At present, the service does not need to actively shut down the http server.

ShiKaiWi (approved):
LGTM

PR #103, opened 2022-11-21
chore: add env variable config parser

ZuLiangWang:

Closes #

# Rationale for this change
At present, the server startup parameters do not support input through environment variables. We need to write some configurations in the configuration file, which is very inflexible. We hope to directly inject parameters through environment variables.

* Add env variables parser.

The user can now configure the startup parameters of the server through the environment variables.

# How does this change test
Local manual test.


chunshao90 (approved):
LGTM

PR #105, opened 2022-11-23
fix: remove dropTable rollback when createTable failed

chunshao90:

Closes #

# Rationale for this change
There is a very serious bug when creating an existing table will cause the old table to be deleted.
 

Remove `dropTable` rollback when `createTable` failed.

No.

# How does this change test
Manual testing.

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -128,7 +128,7 @@ func (c *Cluster) DropTable(ctx context.Context, schemaName, tableName string) (
 	ret := DropTableResult{
 		ShardVersionUpdate: updateVersion,
 	}
-	log.Info(fmt.Sprintf("create table success, result:%v", ret), zap.String("cluster", c.Name()), zap.String("schemaName", schemaName), zap.String("tableName", tableName))
+	log.Info("drop table success", zap.String("cluster", c.Name()), zap.String("schemaName", schemaName), zap.String("tableName", tableName), zap.String("result", fmt.Sprintf("%+v", ret)))
 	return ret, nil
 }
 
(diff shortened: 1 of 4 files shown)


server/cluster/cluster.go
```diff
@@ -122,9 +125,11 @@ func (c *Cluster) DropTable(ctx context.Context, schemaName, tableName string) (
 		return DropTableResult{}, errors.WithMessagef(err, "topology manager remove table")
 	}
 
-	return DropTableResult{
+	ret := DropTableResult{
 		ShardVersionUpdate: updateVersion,
-	}, nil
+	}
+	log.Info(fmt.Sprintf("create table success, result:%v", ret), zap.String("cluster", c.Name()), zap.String("schemaName", schemaName), zap.String("tableName", tableName))
```
ZuLiangWang, line 131:
```suggestion
	log.Info(fmt.Sprintf("drop table success, result:%v", ret), zap.String("cluster", c.Name()), zap.String("schemaName", schemaName), zap.String("tableName", tableName))
```

server/cluster/cluster.go
```diff
@@ -159,10 +166,12 @@ func (c *Cluster) CreateTable(ctx context.Context, nodeName string, schemaName s
 		return CreateTableResult{}, errors.WithMessage(err, "topology manager add table")
 	}
 
-	return CreateTableResult{
+	ret := CreateTableResult{
 		Table:              table,
 		ShardVersionUpdate: result,
-	}, nil
+	}
+	log.Info(fmt.Sprintf("create table succeed, result:%v", ret), zap.String("cluster", c.Name()))
```
ZuLiangWang, line 173: It seems strange to use two output methods at the same time, maybe it is better to use the same.

ZuLiangWang (changes requested):

PR #106, opened 2022-11-23
fix: fix load schema table mapping & logger initialization

ZuLiangWang:

Closes #

# Rationale for this change
When server restart, it will reload all metadata in memory, but schema table mapping is not initialized normally.

* Fix schema table mapping initialization.
* Fix the initialization problem of the logger to avoid invalid log configuration.

None.

# How does this change test
Local manual test.


chunshao90 (approved):
LGTM

PR #108, opened 2022-11-24
fix: fix shard view copy table

ZuLiangWang:
Closes #

# Rationale for this change
Fix the bug that tableID cannot be added normally when creating a table.

* Replace `copy` method with `append`.

None.

# How does this change test
Add multi tables create test. @chunshao90


chunshao90 (approved):
LGTM

PR #112, opened 2022-11-26
feat: parse config from env

ZuLiangWang:

Closes #

# Rationale for this change
The special character '-' is not allowed in the environment variable, so the previous reading method of the environment variable needs to be adjusted.

* Import env parse lib, ref: https://github.com/caarlos0/env

The server could be started through the environment variable.

# How does this change test
Local manual test.


server/config/config.go
```diff
@@ -289,18 +290,9 @@ func (p *Parser) ParseConfigFromToml() error {
 }
 
 func (p *Parser) ParseConfigFromEnvVariables() error {
```
chunshao90, line 292:
```suggestion
func (p *Parser) ParseConfigFromEnv() error {
```

PR #113, opened 2022-11-27
chore: add some debug info when cluster load metadata

chunshao90:

Closes #

# Rationale for this change
Add some debugging information when the cluster loads metadata to find troubleshooting of potential metadata errors.
 

Add some debug info when cluster load metadata.

No.

# How does this change test
No need to test.


ZuLiangWang (approved):
LGTM

PR #114, opened 2022-11-29
refactor: update pb definition

ZuLiangWang:

Closes #

# Rationale for this change
Update the pb dependency in ceresmeta to be consistent with the ceresproto version.

* Update ceresproto version.
* Refactor types definition, keep same with ceresproto.

None.

# How does this change test
Pass all unit tests.


chunshao90 (approved):
LGTM

PR #116, opened 2022-12-05
feat: support forwarding http requests

ZuLiangWang:

Closes #

# Rationale for this change
When CeresMeta is started in the cluster mode, the cluster contains the leader node and follower node, and only the leader node can process requests normally. However, the current http requests sent to the follower node are not forwarded like grpc requests, which will result in that the requests cannot be processed normally.

* Add http forward client implementation.
* Add forward logic, follower node will forwarded request to leader node.

None.

# How does this change test

### DIFF
--- a/server/service/http/forward.go
+++ b/server/service/http/forward.go
@@ -51,19 +51,21 @@ func (s *ForwardClient) getForwardedAddr(ctx context.Context) (string, bool, err
 	if member.IsLocal {
 		return "", true, nil
 	}
-	leaderAddr := strings.Split(member.Leader.GetEndpoint(), ":")
-	leaderAddr[2] = strconv.Itoa(s.port)
-	return strings.Join(leaderAddr, ":"), false, nil
+	httpAddr, err := formatHTTPAddr(member.Leader.Endpoint, s.port)
+	if err != nil {
+		return "", false, errors.WithMessage(err, "format http addr")
+	}
+	return httpAddr, false, nil
 }
 
 func (s *ForwardClient) forwardToLeader(req *http.Request) (*http.Response, bool, error) {
 	addr, isLeader, err := s.getForwardedAddr(req.Context())
 	if err != nil {
 		log.Error("get forward addr failed", zap.Error(err))
-		return &http.Response{}, false, err
+		return nil, false, err
 	}
 	if isLeader {
-		return &http.Response{}, true, nil
+		return nil, true, nil
 	}
 
 	// Update remote host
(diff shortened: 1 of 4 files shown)


server/service/http/forward.go
```diff
@@ -0,0 +1,85 @@
...
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			TLSHandshakeTimeout: 10 * time.Second,
+		},
+	}
+}
+
+func (s *ForwardClient) getForwardedAddr(ctx context.Context) (string, bool, error) {
```
chunshao90, line 46: There is some duplication of code consistent with the forwarding logic in grpc. Define some commonly used methods to avoid duplication of code.

ZuLiangWang: The duplicated code is `member.GetLeader` and valid logic of it's result, I think it is no need to add common method for it.

server/service/http/forward.go
```diff
@@ -0,0 +1,93 @@
...
+		return &http.Response{}, false, err
+	}
+
+	return resp, false, nil
+}
+
+// formatHttpAddr convert grpcAddr(127.0.0.1:8831) httpPort(5000) to httpAddr(127.0.0.1:5000).
+func formatHTTPAddr(grpcAddr string, httpPort int) string {
+	leaderAddr := strings.Split(grpcAddr, ":")
+	leaderAddr[2] = strconv.Itoa(httpPort)
```
chunshao90, line 89: Check leaderAddr len.

server/service/http/forward.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	httpAddr := formatHTTPAddr(member.Leader.Endpoint, s.port)
+	return httpAddr, false, nil
+}
+
+func (s *ForwardClient) forwardToLeader(req *http.Request) (*http.Response, bool, error) {
+	addr, isLeader, err := s.getForwardedAddr(req.Context())
+	if err != nil {
+		log.Error("get forward addr failed", zap.Error(err))
+		return &http.Response{}, false, err
```
chunshao90, line 62:
```suggestion
		return nil, false, err
```

server/service/http/forward.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	if isLeader {
+		return &http.Response{}, true, nil
```
chunshao90, line 65:
```suggestion
		return nil, true, nil
```

chunshao90 (approved):
LGTM

PR #117, opened 2022-12-06
chore: add start config file

ZuLiangWang:

Closes #

# Rationale for this change
In this release, we updated the input mode of the server startup configuration. Now we use the configuration file method to input the startup parameters, so we need to supplement the configuration file example.

* Add server start config examples, including standalone mode and cluster mode.

User could start server with config examples.

# How does this change test
Start normally using the configuration file in the example.


chunshao90 (approved):
LGTM

PR #120, opened 2022-12-09
feat: add etcd max txn ops config

ZuLiangWang:

Closes #

# Rationale for this change
Because of the limitation of etcd on single txn ops (128 by default), when the cluster is configured with a large number of shards, etcd will fail to write. Therefore, it is necessary to provide a configuration entry to adjust the maxTxnOps.

* Add EtcdMaxTxnOps in start config.
* Start etcd server with EtcdMaxTxnOps.

None.

# How does this change test


chunshao90 (approved):
LGTM

PR #123, opened 2022-12-19
feat: add create partition table procedure

ZuLiangWang:

(https://github.com/CeresDB/ceresdb/issues/492)

# Rationale for this change
Support create partition table in CeresDB.

* Add implementation of `CreatePartitionTableProcedure`.
* Add `ShardPicker` for picking up the shards suitable for scheduling in the cluster.

None.

# How does this change test
Pass unit test.

### DIFF
--- a/server/cluster/manager.go
+++ b/server/cluster/manager.go
@@ -103,13 +103,12 @@ func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, opt
 	}
 
 	clusterMetadata := storage.Cluster{
-		ID:                         clusterID,
-		Name:                       clusterName,
-		MinNodeCount:               opts.NodeCount,
-		ReplicationFactor:          opts.ReplicationFactor,
-		ShardTotal:                 opts.ShardTotal,
-		PartitionTableRatioOfNodes: opts.PartitionTableRatioOfNodes,
-		CreatedAt:                  uint64(time.Now().UnixMilli()),
+		ID:                clusterID,
+		Name:              clusterName,
+		MinNodeCount:      opts.NodeCount,
+		ReplicationFactor: opts.ReplicationFactor,
+		ShardTotal:        opts.ShardTotal,
+		CreatedAt:         uint64(time.Now().UnixMilli()),
 	}
 	err = m.storage.CreateCluster(ctx, storage.CreateClusterRequest{
 		Cluster: clusterMetadata,
--- a/server/cluster/manager.go
+++ b/server/cluster/manager.go
@@ -29,7 +29,7 @@ type Manager interface {
 	Stop(ctx context.Context) error
 
 	ListClusters(ctx context.Context) ([]*Cluster, error)
-	CreateCluster(ctx context.Context, clusterName string, nodeCount, replicationFactor, shardTotal, partitionTableNum uint32) (*Cluster, error)
+	CreateCluster(ctx context.Context, clusterName string, opts CreateClusterOpts) (*Cluster, error)
 	GetCluster(ctx context.Context, clusterName string) (*Cluster, error)
 	// AllocSchemaID means get or create schema.
 	// The second output parameter bool: Returns true if the table was newly created.
@@ -82,10 +82,8 @@ func (m *managerImpl) ListClusters(_ context.Context) ([]*Cluster, error) {
 	return clusters, nil
 }
 
-func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, initialNodeCount,
-	replicationFactor, shardTotal, partitionTableNum uint32,
-) (*Cluster, error) {
-	if initialNodeCount < 1 {
+func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, opts CreateClusterOpts) (*Cluster, error) {
+	if opts.NodeCount < 1 {
 		log.Error("cluster's nodeCount must > 0", zap.String("clusterName", clusterName))
 		return nil, ErrCreateCluster.WithCausef("nodeCount must > 0")
 	}
@@ -107,10 +105,10 @@ func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, ini
 	clusterMetadata := storage.Cluster{
 		ID:                clusterID,
 		Name:              clusterName,
-		MinNodeCount:      initialNodeCount,
-		ReplicationFactor: replicationFactor,
-		ShardTotal:        shardTotal,
-		PartitionTableNum: partitionTableNum,
+		MinNodeCount:      opts.NodeCount,
+		ReplicationFactor: opts.ReplicationFactor,
+		ShardTotal:        opts.ShardTotal,
+		TablePartitionNum: opts.TablePartitionNum,
 		CreatedAt:         uint64(time.Now().UnixMilli()),
 	}
 	err = m.storage.CreateCluster(ctx, storage.CreateClusterRequest{
--- a/server/cluster/manager.go
+++ b/server/cluster/manager.go
@@ -103,13 +103,13 @@ func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, opt
 	}
 
 	clusterMetadata := storage.Cluster{
-		ID:                clusterID,
-		Name:              clusterName,
-		MinNodeCount:      opts.NodeCount,
-		ReplicationFactor: opts.ReplicationFactor,
-		ShardTotal:        opts.ShardTotal,
-		TablePartitionNum: opts.TablePartitionNum,
-		CreatedAt:         uint64(time.Now().UnixMilli()),
+		ID:                         clusterID,
+		Name:                       clusterName,
+		MinNodeCount:               opts.NodeCount,
+		ReplicationFactor:          opts.ReplicationFactor,
+		ShardTotal:                 opts.ShardTotal,
+		PartitionTableRatioOfNodes: opts.PartitionTableRatioOfNodes,
+		CreatedAt:                  uint64(time.Now().UnixMilli()),
 	}
 	err = m.storage.CreateCluster(ctx, storage.CreateClusterRequest{
 		Cluster: clusterMetadata,
--- a/server/cluster/manager.go
+++ b/server/cluster/manager.go
@@ -29,7 +29,7 @@ type Manager interface {
 	Stop(ctx context.Context) error
 
 	ListClusters(ctx context.Context) ([]*Cluster, error)
-	CreateCluster(ctx context.Context, clusterName string, nodeCount, replicationFactor, shardTotal uint32) (*Cluster, error)
+	CreateCluster(ctx context.Context, clusterName string, nodeCount, replicationFactor, shardTotal, partitionTableNum uint32) (*Cluster, error)
 	GetCluster(ctx context.Context, clusterName string) (*Cluster, error)
 	// AllocSchemaID means get or create schema.
 	// The second output parameter bool: Returns true if the table was newly created.
@@ -83,7 +83,7 @@ func (m *managerImpl) ListClusters(_ context.Context) ([]*Cluster, error) {
 }
 
 func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, initialNodeCount,
-	replicationFactor, shardTotal uint32,
+	replicationFactor, shardTotal, partitionTableNum uint32,
 ) (*Cluster, error) {
 	if initialNodeCount < 1 {
 		log.Error("cluster's nodeCount must > 0", zap.String("clusterName", clusterName))
@@ -110,6 +110,7 @@ func (m *managerImpl) CreateCluster(ctx context.Context, clusterName string, ini
 		MinNodeCount:      initialNodeCount,
 		ReplicationFactor: replicationFactor,
 		ShardTotal:        shardTotal,
+		PartitionTableNum: partitionTableNum,
 		CreatedAt:         uint64(time.Now().UnixMilli()),
 	}
 	err = m.storage.CreateCluster(ctx, storage.CreateClusterRequest{
--- a/server/cluster/table_manager_test.go
+++ b/server/cluster/table_manager_test.go
@@ -160,10 +160,9 @@ func testCluster(ctx context.Context, re *require.Assertions, manager Manager, c
 
 func testCreateCluster(ctx context.Context, re *require.Assertions, manager Manager, clusterName string) {
 	_, err := manager.CreateCluster(ctx, clusterName, CreateClusterOpts{
-		NodeCount:                  defaultNodeCount,
-		ReplicationFactor:          defaultReplicationFactor,
-		ShardTotal:                 defaultShardTotal,
-		PartitionTableRatioOfNodes: defaultPartitionTableRatioOfNodes,
+		NodeCount:         defaultNodeCount,
+		ReplicationFactor: defaultReplicationFactor,
+		ShardTotal:        defaultShardTotal,
 	})
 	re.NoError(err)
 }
@@ -194,7 +193,7 @@ func testCreateTable(ctx context.Context, re *require.Assertions, manager Manage
 ) {
 	c, err := manager.GetCluster(ctx, clusterName)
 	re.NoError(err)
-	table, err := c.CreateTable(ctx, node, schema, tableName)
+	table, err := c.CreateTable(ctx, node, schema, tableName, false)
 	re.NoError(err)
 	re.Equal(tableID, table.Table.ID)
 }
--- a/server/cluster/table_manager_test.go
+++ b/server/cluster/table_manager_test.go
@@ -17,29 +17,29 @@ import (
 )
 
 const (
-	defaultTimeout                  = time.Second * 10
-	cluster1                        = "ceresdbCluster1"
-	cluster2                        = "ceresdbCluster2"
-	defaultSchema                   = "ceresdbSchema"
-	defaultNodeCount                = 2
-	defaultReplicationFactor        = 1
-	defaultShardTotal               = 8
-	defaultTablePartitionNum        = 2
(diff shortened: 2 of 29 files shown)


server/cluster/manager.go
```diff
@@ -29,7 +29,7 @@ type Manager interface {
 	Stop(ctx context.Context) error
 
 	ListClusters(ctx context.Context) ([]*Cluster, error)
-	CreateCluster(ctx context.Context, clusterName string, nodeCount, replicationFactor, shardTotal uint32) (*Cluster, error)
+	CreateCluster(ctx context.Context, clusterName string, nodeCount, replicationFactor, shardTotal, partitionTableNum uint32) (*Cluster, error)
```
chunshao90, line 32: Rename `partitionTableNum` to `TablePartitionNum`?

server/coordinator/procedure/ShardPicker.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/CeresDB/ceresmeta/server/cluster"
+	"github.com/pkg/errors"
+)
+
+// ShardPicker is used to pick up the shards suitable for scheduling in the cluster.
+type ShardPicker interface {
+	PickShards(ctx context.Context, clusterName string, num int) ([]cluster.ShardNodeWithVersion, error)
+}
+
+// RandomShardPicker randomly pick up shards not on the same node exists in current cluster.
```
chunshao90, line 20:
```suggestion
// RandomShardPicker randomly pick up shards that are not on the same node in the current cluster.
```

server/storage/types.go
```diff
@@ -139,6 +139,7 @@ type Cluster struct {
 	MinNodeCount      uint32
 	ReplicationFactor uint32
 	ShardTotal        uint32
+	PartitionTableNum uint32
```
chunshao90, line 142: Need to modify Cluster pb in `ceresdbproto`.

server/service/grpc/service.go
```diff
@@ -170,17 +170,33 @@ func (s *Service) CreateTable(ctx context.Context, req *metaservicepb.CreateTabl
...
-		Cluster:     c,
-		SourceReq:   req,
-		OnSucceeded: onSucceeded,
-		OnFailed:    onFailed,
-	})
-	if err != nil {
-		log.Error("fail to create table", zap.Error(err))
-		return &metaservicepb.CreateTableResponse{Header: responseHeader(err, "create table")}, nil
+	var p procedure.Procedure
+	if req.PartitionInfo != nil && len(req.PartitionInfo.GetNames()) != 0 {
```
chunshao90, line 174: `if req.PartitionInfo != nil && len(req.PartitionInfo.GetNames()) == 0`, shall we return error?
Shall we paas `CreateTableRequest` to `factory`, `factory` decides  to create `CreatePartitionTableProcedure` or `CreateTableProcedure`.

server/storage/types.go
ZuLiangWang: Update pb definition of `Cluster` in this pr: https://github.com/CeresDB/ceresdbproto/pull/39/files

server/service/grpc/service.go
ShiKaiWi: I guess it is not good to do validation for `CreateTableRequest` in the service layer because the service layer have no idea about the specific logic about what is necessary to create table and another validation is still required in the procedure layer even if it is done by the service layer.

However, it is not good to let the `factory` to take over such validation because of the same reason: the `factory` should not have any knowledge about that.

So I guess the best way is to let the specific procedure to do such validation, e.g. `CreatePartitionTableProcedure`.

server/service/grpc/service.go
```diff
@@ -170,17 +170,37 @@ func (s *Service) CreateTable(ctx context.Context, req *metaservicepb.CreateTabl
...
-		Cluster:     c,
-		SourceReq:   req,
-		OnSucceeded: onSucceeded,
-		OnFailed:    onFailed,
-	})
-	if err != nil {
-		log.Error("fail to create table", zap.Error(err))
-		return &metaservicepb.CreateTableResponse{Header: responseHeader(err, "create table")}, nil
+	var p procedure.Procedure
+	if req.PartitionInfo != nil && len(req.PartitionInfo.GetNames()) == 0 {
```
ShiKaiWi, line 174: I guess it is not good to do validation for `CreateTableRequest` in the service layer because the service layer have no idea about the specific logic about what is necessary to create table and another validation is still required in the procedure layer even if it is done by the service layer.

However, it is not good to let the `factory` to take over such validation because of the same reason: the `factory` should not have any knowledge about that.

So I guess the best way is to let the specific procedure to do such validation, e.g. `CreateTableProcedure`.

server/service/grpc/service.go
ShiKaiWi: I guess it is not a good idea to decide which procedure to create in the service layer because service layer has no knowledge about how to create table.

I guess we can make this decision in `CreateTableProcedure`, that is to say, the procedure factory has no idea about the `CreatePartitionTableProcedure`, which is just a creating branch in the `CreateTableProcedure`. And there are still two different specific procedures according to the `CreateTableRequest`.

server/coordinator/procedure/ShardPicker.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		nodeShardsMapping[nodeShard.ShardNode.NodeName] = append(nodeShardsMapping[nodeShard.ShardNode.NodeName], nodeShard)
+	}
+	if len(nodeShardsMapping) < num {
+		return []cluster.ShardNodeWithVersion{}, errors.WithMessage(ErrNodeNumberNotEnough, fmt.Sprintf("number of nodes is: %d, expecet number of shards is: %d", len(nodeShardsMapping), num))
+	}
+
+	var selectNodeNames []string
+	for i := 0; i < num; i++ {
+		selectNodeIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(nodeNames))))
```
ShiKaiWi, line 53: Across different iterations, will `selectNodeIndex` be duplicate? Or it is ok to pick duplicate node?

server/coordinator/procedure/create_partition_table.go
```diff
@@ -0,0 +1,342 @@
...
+		cluster:              p.cluster,
+		dispatch:             p.dispatch,
+		sourceReq:            p.req,
+		partitionTableShards: p.partitionTableShards,
+		dataTablesShards:     p.dataTablesShards,
+		onSucceeded:          p.onSucceeded,
+		onFailed:             p.onFailed,
+	}
+
+	for {
```
ShiKaiWi, line 109: Shall we add an ascii diagram to show the state machine?

server/coordinator/procedure/create_partition_table.go
```diff
@@ -0,0 +1,342 @@
...
+	dataTablesShards     []cluster.ShardNodeWithVersion
+
+	onSucceeded func(cluster.CreateTableResult) error
+	onFailed    func(error) error
+
+	lock  sync.RWMutex
+	state State
+}
+
+func NewCreatePartitionTableProcedure(id uint64, cluster *cluster.Cluster, dispatch eventdispatch.Dispatch, storage Storage, req *metaservicepb.CreateTableRequest, partitionTableShards []cluster.ShardNodeWithVersion, dataTablesShards []cluster.ShardNodeWithVersion, onSucceeded func(cluster.CreateTableResult) error, onFailed func(error) error) *CreatePartitionTableProcedure {
```
ShiKaiWi, line 67: The input params are too many, and maybe we should define struct to replace most of these parameters.

server/coordinator/procedure/factory.go
```diff
@@ -131,6 +143,32 @@ func (f *Factory) CreateSplitProcedure(ctx context.Context, request SplitRequest
 	return procedure, nil
 }
 
+func (f *Factory) CreateCreatePartitionTableProcedure(ctx context.Context, request CreatePartitionTableRequest) (Procedure, error) {
```
ShiKaiWi: `CreateCreate...` seems weird, how about `MakeCreate...`?

server/service/grpc/service.go
ZuLiangWang: The content of this conversation is the same as that above, resolve it.

server/cluster/table_manager_test.go
```diff
@@ -39,6 +39,7 @@ const (
 	testRootPath                    = "/rootPath"
 	defaultIDAllocatorStep          = 20
 	defaultThreadNum                = 20
+	defaultTablePartitionNum        = 2
```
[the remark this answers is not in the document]
ZuLiangWang: Yes, it is a cluster option, only the unit test here uses a default value.

server/coordinator/procedure/ShardPicker.go
```diff
@@ -0,0 +1,72 @@
...
+	clusterManager cluster.Manager
+}
+
+func NewRandomShardPicker(manager cluster.Manager) ShardPicker {
+	return &RandomShardPicker{
+		clusterManager: manager,
+	}
+}
+
+func (p *RandomShardPicker) PickShards(ctx context.Context, clusterName string, num int) ([]cluster.ShardNodeWithVersion, error) {
```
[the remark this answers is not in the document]
ZuLiangWang: Replace param name `num` to `expectShardNum` maybe enough?

server/coordinator/procedure/ShardPicker.go
ZuLiangWang:
```
	for i := 0; i < expectShardNum; i++ {
		selectNodeIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(nodeNames))))
		if err != nil {
			return []cluster.ShardNodeWithVersion{}, errors.WithMessage(err, "generate random node index")
		}
		selectNodeNames = append(selectNodeNames, nodeNames[selectNodeIndex.Int64()])
		nodeNames = append(nodeNames[:selectNodeIndex.Int64()], nodeNames[selectNodeIndex.Int64()+1:]...)
	}
```
The select node will be removed from `NodeNames`, so it will not pick duplicate nodes.

server/cluster/table_manager_test.go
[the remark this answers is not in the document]
ShiKaiWi: I guess different tables should have different partition number which is configured by the user's create table request.

server/coordinator/procedure/ShardPicker.go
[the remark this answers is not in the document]
ShiKaiWi: Such a renaming is good, and comments are also necessary. Actually, I hope most of exposed method or struct should be commented for their usage.

server/coordinator/procedure/ShardPicker.go
ShiKaiWi: I see.
However, this seems very inefficient because it leads too many copies. Maybe you can swap the selected node with the last one and use a new length = (previous length - 1) for next iteration, instead of such copy removing.

server/cluster/table_manager_test.go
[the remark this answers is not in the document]
ZuLiangWang: User specifies the number of partitions in the partition table through SQL like this:
```
CREATE TABLE `demo` (`name` string TAG, `value` double NOT NULL, `t` timestamp NOT NULL, TIMESTAMP KEY(t)) PARTITION BY KEY(name) PARTITIONS 2 ENGINE=Analytic with (enable_ttl="false")
```

`TablePartitionNum` is not the number of partitions in the partition table, but the number of routing nodes in the partition, which is transparent to users.

server/coordinator/procedure/ShardPicker.go
```diff
@@ -0,0 +1,72 @@
...
+	getNodeShardResult, err := p.clusterManager.GetNodeShards(ctx, clusterName)
```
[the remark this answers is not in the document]
ZuLiangWang: We can directly reuse the public methods of previous unit tests. It is not troublesome to construct a shardPicker, so we do not need to make this change separately for testing.

server/service/grpc/service.go
ZuLiangWang: Solve this problem by adding a proxy procedure.

server/cluster/table_manager_test.go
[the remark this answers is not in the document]
ShiKaiWi: The name is confusing...

Maybe this should not be configured as an option of cluster? It can just be a default value, and I guess a better way is to define a ratio based on the total nodes telling us to choose how many routing nodes to select for partition table.

server/coordinator/procedure/create_table.go
```diff
@@ -4,215 +4,71 @@ package procedure
...
+		log.Error("fail to create table", zap.Error(ErrEmptyPartitionNames))
+		return CreateTableProcedure{}, ErrEmptyPartitionNames
 	}
-
-	if err := p.fsm.Event(eventCreateTablePrepare, req); err != nil {
-		err1 := p.fsm.Event(eventCreateTableFailed, req)
-		p.updateState(StateFailed)
-		if err1 != nil {
-			err = errors.WithMessagef(err, "send eventCreateTableFailed, err:%v", err1)
+	var realProcedure Procedure
```
ShiKaiWi, line 24: Maybe one new line here is better.

server/coordinator/procedure/create_table.go
```diff
@@ -4,215 +4,71 @@ package procedure
...
+		if err != nil {
+			log.Error("fail to create table", zap.Error(err))
+			return CreateTableProcedure{}, err
+		}
+		realProcedure = p
 	}
-
-	p.updateState(StateFinished)
-	return nil
+	return CreateTableProcedure{
```
ShiKaiWi, line 51: One newline above seems better.

server/storage/types.go
ShiKaiWi: Let's remove it from cluster. It should not be a cluster option.

server/cluster/types.go
```diff
@@ -35,6 +35,13 @@ type ShardNodeWithVersion struct {
 	ShardNode storage.ShardNode
 }
 
+type CreateClusterOpts struct {
+	NodeCount         uint32
+	ReplicationFactor uint32
+	ShardTotal        uint32
+	PartitionRadio    uint32
```
ShiKaiWi, line 42: Maybe this is a better name `PartitionTableRatioOfNodes`?

server/cluster/table_manager_test.go
```diff
@@ -17,28 +17,29 @@ import (
...
-	defaultIDAllocatorStep          = 20
-	defaultThreadNum                = 20
+	defaultTimeout                           = time.Second * 10
+	cluster1                                 = "ceresdbCluster1"
+	cluster2                                 = "ceresdbCluster2"
+	defaultSchema                            = "ceresdbSchema"
+	defaultNodeCount                         = 2
+	defaultReplicationFactor                 = 1
+	defaultShardTotal                        = 8
+	defaultPartitionTableRatioOfNodes        = 2
```
ShiKaiWi, line 27: This should be a float?

server/config/config.go
```diff
@@ -43,10 +43,11 @@ const (
...
 
-	defaultClusterName              = "defaultCluster"
-	defaultClusterNodeCount         = 2
-	defaultClusterReplicationFactor = 1
-	defaultClusterShardTotal        = 8
+	defaultClusterName                       = "defaultCluster"
+	defaultClusterNodeCount                  = 2
+	defaultClusterReplicationFactor          = 1
+	defaultClusterShardTotal                 = 8
+	defaultClusterPartitionTableRatioOfNodes = 2
```
ShiKaiWi, line 50: The type should be float?

ZuLiangWang: I think int val maybe enough, since the type of nodes numbers is int, is it enough for us to use int radio directly? The use of float will not improve the accuracy.

server/coordinator/procedure/common_test.go
```diff
@@ -16,17 +16,18 @@ import (
...
+	testTableName1                    = "table1"
+	testSchemaName                    = "testSchemaName"
+	nodeName0                         = "node0"
+	nodeName1                         = "node1"
+	testRootPath                      = "/rootPath"
+	defaultIDAllocatorStep            = 20
+	clusterName                       = "ceresdbCluster1"
+	defaultNodeCount                  = 2
+	defaultReplicationFactor          = 1
+	defaultPartitionTableRatioOfNodes = 1
```
ShiKaiWi, line 29: Type seems wrong, and why do the wrong type pass the check?

server/config/config.go
```diff
@@ -43,10 +43,11 @@ const (
...
 
-	defaultClusterName              = "defaultCluster"
-	defaultClusterNodeCount         = 2
-	defaultClusterReplicationFactor = 1
-	defaultClusterShardTotal        = 8
+	defaultClusterName                     = "defaultCluster"
+	defaultClusterNodeCount                = 2
+	defaultClusterReplicationFactor        = 1
+	defaultClusterShardTotal               = 8
+	defaultPartitionTableProportionOfNodes = float32(0.5)
```
ShiKaiWi, line 50: I guess no need to do explicit conversion:
```suggestion
	defaultPartitionTableProportionOfNodes = 0.5
```

Remember to fix the same problems at other places.

server/coordinator/procedure/factory.go
```diff
@@ -20,6 +20,10 @@ type Factory struct {
 	dispatch       eventdispatch.Dispatch
 	storage        Storage
 	clusterManager cluster.Manager
+	shardPicker    ShardPicker
+
+	// TODO: This is a temporary implementation version, which needs to be refined to the table level later.
+	PartitionTableProportionOfNodes float32
```
ShiKaiWi, line 26: Seems no need to expose this field?

ShiKaiWi (approved):
LGTM

PR #131, opened 2023-01-15
feat: store partition info in ceresmeta

chunshao90:

Closes #

# Rationale for this change
 In order to support the case where the partition table is opened by multiple machines in a non-cloud environment, store the partition info of `PartitionTable` in `ceresmeta`.
 

* Store partition info of `PartitionInfo` in metadata of the table.

No.

# How does this change test
Manual testing.

### DIFF
--- a/server/cluster/table_manager_test.go
+++ b/server/cluster/table_manager_test.go
@@ -191,7 +191,12 @@ func testCreateTable(ctx context.Context, re *require.Assertions, manager Manage
 ) {
 	c, err := manager.GetCluster(ctx, clusterName)
 	re.NoError(err)
-	table, err := c.CreateTable(ctx, shardID, schema, tableName, false)
+	table, err := c.CreateTable(ctx, CreateTableRequest{
+		ShardID:       shardID,
+		SchemaName:    schema,
+		TableName:     tableName,
+		PartitionInfo: storage.PartitionInfo{},
+	})
 	re.NoError(err)
 	re.Equal(tableID, table.Table.ID)
 }
--- a/server/coordinator/procedure/create_drop_partition_table_test.go
+++ b/server/coordinator/procedure/create_drop_partition_table_test.go
@@ -7,13 +7,16 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/CeresDB/ceresdbproto/golang/pkg/clusterpb"
 	"github.com/CeresDB/ceresdbproto/golang/pkg/metaservicepb"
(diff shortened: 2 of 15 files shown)


server/cluster/table_manager_test.go
```diff
@@ -191,7 +191,12 @@ func testCreateTable(ctx context.Context, re *require.Assertions, manager Manage
 ) {
 	c, err := manager.GetCluster(ctx, clusterName)
 	re.NoError(err)
-	table, err := c.CreateTable(ctx, shardID, schema, tableName, false)
+	table, err := c.CreateTable(ctx, CreateTableRequest{
+		ShardID:       shardID,
+		SchemaName:    schema,
+		TableName:     tableName,
+		PartitionInfo: storage.PartitionInfo{},
```
ZuLiangWang, line 198: Remove it, `table_manager_test` used to test normal table, `PartitionInfo` should be nil.
If you want to test create partition table in cluster, you should define the new function`testCreatePartitionTable`.

server/coordinator/procedure/create_drop_partition_table_test.go
```diff
@@ -21,7 +21,7 @@ func TestCreateAndDropPartitionTable(t *testing.T) {
 
 	shardPicker := NewRandomBalancedShardPicker(manager)
 
-	testTableNum := 20
+	testTableNum := 8
```
ZuLiangWang, line 24: Why change the number of tables?

server/cluster/table_manager_test.go
chunshao90: `storage.PartitionInfo{}` means the table is not a `PartitionTable`.

server/coordinator/procedure/create_drop_partition_table_test.go
chunshao90: There is no need to create so many partition tables, the current value has already covered the required path.

server/storage/types.go
```diff
@@ -163,12 +163,20 @@ type Schema struct {
...
-	Partitioned bool
+	ID            TableID
+	Name          string
+	SchemaID      SchemaID
+	CreatedAt     uint64
+	PartitionInfo PartitionInfo
+}
+
+func (t Table) IsPartitioned() bool {
+	return t.PartitionInfo.Info != nil
```
ZuLiangWang, line 179: Add check for `t.PartitionInfo`.

ZuLiangWang (approved):
LGTM

Issue #145, opened 2023-03-21
Tracking issue: refactor cluster mode


ZuLiangWang:
**Description**
The cluster mode of CeresDB has many defects at present, and we need a better distributed solution. We also need to refactor the current code to make the cluster mode of CeresDB more robust.

**Proposal**
- [ ] Support shard to transfer leader based on distributed lock using `etcd`. The modifications to the CeresDB section are in this pull request. https://github.com/CeresDB/ceresdb/pull/706
- [ ] Refactor scheduler module,  based on the new interaction model, use scheduler to manage the creation of procedures.
- [ ] Refactor procedure manager, support executing procedures in shard-level serial manner. (Need to support scenarios where one procedure is associated with multiple shards.) 
- [ ] Refactor all types of procedures to remove internal modifications to shard metadata and instead update them through heartbeat.

**Additional context**

PR #146, opened 2023-03-22
refactor: implement failover based distributed etcd lock

ZuLiangWang:

# Rationale for this change
 In order to ensure that user data in distributed mode will not be lost, we need a mechanism to ensure that CeresDB will not have multiple leader shard under any circumstances. `etcd` lock has been added in https://github.com/CeresDB/ceresdb/pull/706 , we refactor scheduler to implement failover based `etcd` distributed lock.

* Add shard watcher to get notify when shard lock is expired.

None.

# How does this change test

### DIFF
--- a/server/coordinator/watch/watch.go
+++ b/server/coordinator/watch/watch.go
@@ -1,0 +1,131 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+package coordinator
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/CeresDB/ceresmeta/pkg/log"
+	"github.com/CeresDB/ceresmeta/server/storage"
+	"github.com/pkg/errors"
+	"go.etcd.io/etcd/api/v3/mvccpb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
+)
+
+type ShardEventType uint64
+
+const (
+	shardPath                  = "shards"
+	EventDelete ShardEventType = 0
+	EventPut                   = 1
+)
+
+// ShardWatch used to watch the distributed lock of shard, and provide the corresponding callback function.
+type ShardWatch struct {
+	rootPath       string
+	etcdClient     *clientv3.Client
+	eventCallbacks []func(shardID storage.ShardID, nodeName string, eventType ShardEventType) error
+
+	lock      sync.RWMutex
+	isRunning bool
+	quit      chan bool
+}
+
+func NewWatch(rootPath string, client *clientv3.Client) *ShardWatch {
+	return &ShardWatch{
+		rootPath:       rootPath,
+		etcdClient:     client,
+		eventCallbacks: []func(shardID storage.ShardID, nodeName string, eventType ShardEventType) error{},
+		quit:           make(chan bool),
+	}
+}
+
+func (w *ShardWatch) Start(ctx context.Context) error {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	if w.isRunning {
+		return nil
+	}
+
+	shardsKeyPath := strings.Join([]string{w.rootPath, shardPath}, "/")
+
+	if err := w.startWatch(ctx, shardsKeyPath); err != nil {
+		return errors.WithMessage(err, "etcd register watch failed")
+	}
+
+	w.isRunning = true
+	return nil
+}
+
+func (w *ShardWatch) Stop(ctx context.Context) error {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	w.isRunning = false
+	w.quit <- true
+	return nil
+}
+
+func (w *ShardWatch) RegisteringEventCallback(eventCallback func(shardID storage.ShardID, nodeName string, eventType ShardEventType) error) {
+	w.eventCallbacks = append(w.eventCallbacks, eventCallback)
+}
+
+func (w *ShardWatch) startWatch(ctx context.Context, path string) error {
+	log.Info("register shard watch", zap.String("watchPath", path))
+	go func() {
+		for {
+			select {
+			case <-w.quit:
+				return
+			default:
+				respChan := w.etcdClient.Watch(ctx, path, clientv3.WithPrefix(), clientv3.WithPrevKV())
+				for resp := range respChan {
+					for _, event := range resp.Events {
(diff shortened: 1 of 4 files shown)


server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,131 @@
...
+	shardPath                  = "shards"
+	EventDelete ShardEventType = 0
+	EventPut                   = 1
+)
+
+// ShardWatch used to watch the distributed lock of shard, and provide the corresponding callback function.
+type ShardWatch struct {
+	rootPath       string
+	etcdClient     *clientv3.Client
+	eventCallbacks []func(shardID storage.ShardID, nodeName string, eventType ShardEventType) error
```
ShiKaiWi, line 32: Maybe we should define an interface for the callbacks for explicit meaning of its parameter:
```
type ShardEventCallback interface {
   OnShardRegistered(ShardRegisterEvent) error
   OnShardExpired(ShardExpireEvent) error
}
```

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,131 @@
...
+
+	lock      sync.RWMutex
+	isRunning bool
+	quit      chan bool
```
ShiKaiWi, line 36:
```suggestion
	quit      chan struct{}
```

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (w *ShardWatch) processEvent(event *clientv3.Event) error {
+	switch event.Type {
+	case mvccpb.DELETE:
+		pathList := strings.Split(string(event.Kv.Key), "/")
+		shardID, err := strconv.ParseUint(pathList[len(pathList)-1], 10, 64)
+		if err != nil {
+			return err
+		}
```
ShiKaiWi, line 108: Add methods `encodeKey` and `decodeKey` to replace the duplicate manipulations. And wrap the error because it should a leaf error.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (w *ShardWatch) processEvent(event *clientv3.Event) error {
+	switch event.Type {
+	case mvccpb.DELETE:
+		shardID, err := decodeKey(event)
+		if err != nil {
+			return err
+		}
+		oldLeader := string(event.PrevKv.Value)
```
ShiKaiWi, line 119: The content of the value maybe should be defined as a protobuf message, which will be shared with ceresdb.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (w *ShardWatch) Start(ctx context.Context) error {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	if w.isRunning {
+		return nil
+	}
+
+	shardsKeyPath := strings.Join([]string{w.rootPath, shardPath}, "/")
```
ShiKaiWi, line 67: Define a method called `encodeKey` similar to the `decodeKey`.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,155 @@
...
+			default:
+				respChan := w.etcdClient.Watch(ctx, path, clientv3.WithPrefix(), clientv3.WithPrevKV())
+				for resp := range respChan {
+					for _, event := range resp.Events {
+						if err := w.processEvent(event); err != nil {
+							log.Error("process event", zap.Error(err))
+						}
+					}
+				}
+			}
```
ShiKaiWi, line 106: Is it able to quit when `quit` message is received during the watching procedure.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,155 @@
...
+				NewLeaderNode: newLeader,
+			}); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func decodeKey(event *clientv3.Event) (uint64, error) {
```
ShiKaiWi, line 148: It should only receive a string `rawKey` as a parameter.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,155 @@
...
+	pathList := strings.Split(string(event.Kv.Key), "/")
```
ShiKaiWi, line 149: Define a constant for "/" will be used when encode the key.

server/coordinator/watch/watch.go
ShiKaiWi: I've left some comments for the proto pr. You'd better to address them first.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,154 @@
...
+func decodeKey(keyPath string) (uint64, error) {
+	pathList := strings.Split(keyPath, "/")
+	shardID, err := strconv.ParseUint(pathList[len(pathList)-1], 10, 64)
+	if err != nil {
+		return 0, errors.WithMessage(err, "decode etcd event key failed")
+	}
+	return shardID, nil
+}
+
+func encodeKey(rootPath string, shardPath string, shardID uint64) string {
```
ShiKaiWi, line 152:
```suggestion
func encodeKey(rootPath, shardPath string, shardID uint64) string {
```

BTW, this method seems used only in test codes. How about re-implementing it as `encodeShardKeyPrefix` which can be used at L67.

And rename the `decodeKey` to `decodeShardKey`.

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,154 @@
...
+			}); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func decodeKey(keyPath string) (uint64, error) {
+	pathList := strings.Split(keyPath, "/")
```
ShiKaiWi, line 144:
```suggestion
	pathList := strings.Split(keyPath, keySep)
```

server/coordinator/watch/watch.go
```diff
@@ -0,0 +1,174 @@
+// Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
+
+package coordinator
+
+import (
+	"context"
+	"fmt"
+	"github.com/CeresDB/ceresdbproto/golang/pkg/metaeventpb"
+	"google.golang.org/protobuf/proto"
```
ShiKaiWi, line 9: Move them to L14

ShiKaiWi (approved):

PR #147, opened 2023-03-22
refactor: refactor scheduler module

ZuLiangWang:

https://github.com/CeresDB/ceresmeta/issues/145

# Rationale for this change
Refactor scheduler module with reference to the implementation of PD, and use Scheduler as a scheduling generator for Procedure.

1. Add `schedulerManager` to register new scheduler and manage it.
2. Implement `AssignShardScheduler` based on `schedulerManager`.

None.

# How does this change test


server/coordinator/node_picker.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+	"crypto/rand"
+	"math/big"
+
+	"github.com/CeresDB/ceresmeta/server/cluster"
+	"github.com/pkg/errors"
+)
+
+type NodePicker interface {
+	PickNode(ctx context.Context, clusterName string) (cluster.RegisteredNode, error)
```
ShiKaiWi, line 15: The `NodePicker` doesn't have to know anything about the cluster, so how about put it as:
```suggestion
	PickNode(ctx context.Context, nodes []*cluster.RegisteredNode) (int, error)
```

After that, unit tests will be more simple.

server/coordinator/node_picker.go
```diff
@@ -0,0 +1,42 @@
...
+	for _, node := range nodes {
+		if !node.IsOnline() {
+			onlineNodes = append(onlineNodes, node)
+		}
+	}
+
+	selectNodeIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(onlineNodes))))
+	if err != nil {
+		return cluster.RegisteredNode{}, errors.WithMessage(err, "generate random node index")
+	}
```
ShiKaiWi, line 39: Actually, we can avoid the slice creation:
```go
numOnlineNodes := 0
for idx := range nodes {
   if nodes[idx].IsOnline() {
       numOnlineNodes ++
   }
}
selectedIdx, err := rand.Int(rand.Reader, big.NewInt(numOnlineNodes))
for {
    if selectedIdx >= len(nodes) {
       // No valid node is found.
        return -1, nil
    }
    
    if nodes[selectedIdx].IsOnline() {
        return selectedIdx
    }
    selectedIdx++
}
```

server/coordinator/procedure/operation/split/split.go
```diff
@@ -215,7 +215,7 @@ func openNewShardMetadataCallback(event *fsm.Event) {
 	}
 
 	// Validate tables.
-	shardTables := req.cluster.GetShardTables([]storage.ShardID{req.shardID}, req.targetNodeName)[req.shardID]
+	shardTables := req.cluster.GetShardTables([]storage.ShardID{req.shardID})[req.shardID]
```
ShiKaiWi, line 218: I guess it is possible to be nil?

server/coordinator/scheduler/scheduler.go
```diff
@@ -0,0 +1,21 @@
...
+
+type Result struct {
+	p procedure.Procedure
+	// Scheduler will give the reason than why the procedure is generated.
+	schedulerReason string
+}
+
+type Scheduler interface {
+	// Schedule will generate procedure based on current cluster topology, which will be submitted to ProcedureManager,and whether it is actually executed depends on the current state of ProcedureManager.
+	Schedule(ctx context.Context, clusterView storage.ClusterView, shardViews []storage.ShardView) (error, Result)
```
ShiKaiWi:
```suggestion
	Schedule(ctx context.Context, clusterView storage.ClusterView, shardViews []storage.ShardView) (Result, error)
```

server/coordinator/scheduler/scheduler.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/CeresDB/ceresmeta/server/coordinator/procedure"
+	"github.com/CeresDB/ceresmeta/server/storage"
+)
+
+type Result struct {
+	p procedure.Procedure
+	// Scheduler will give the reason than why the procedure is generated.
+	schedulerReason string
+}
```
ShiKaiWi, line 16:
```suggestion
type ScheduleResult struct {
	P procedure.Procedure
	// Scheduler will give the reason than why the procedure is generated.
	Reason string
}
```

server/coordinator/scheduler/scheduler.go
```diff
@@ -0,0 +1,21 @@
...
+
+type Scheduler interface {
+	// Schedule will generate procedure based on current cluster topology, which will be submitted to ProcedureManager,and whether it is actually executed depends on the current state of ProcedureManager.
```
ShiKaiWi, line 19:
```suggestion
	// Scheduler will generate procedure based on current cluster topology, which will be submitted to ProcedureManager, and whether it is actually executed depends on the current state of ProcedureManager.
```

server/server.go
```diff
@@ -48,8 +49,8 @@ type Server struct {
 	procedureFactory *coordinator.Factory
 	// procedureManager process procedure for all cluster.
 	procedureManager procedure.Manager
-	// scheduler schedule procedure for all cluster.
-	scheduler *coordinator.Scheduler
+	// schedulerManager manager schedulers for all cluster.
```
ShiKaiWi, line 52:
```suggestion
	// schedulerManager manages schedulers for all cluster.
```

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/CeresDB/ceresmeta/server/storage"
+	"go.uber.org/zap"
+)
+
+const (
+	schedulerInterval = time.Second * 5
+)
+
+// Manager used to manage schedulers.
+// Each registered scheduler will generate a procedure if the cluster s
```
ShiKaiWi, line 25: The comments seem not integrate.
```suggestion
// Manager manages all schedulers.
//
// Each registered scheduler will generate a procedure if the clusters ...
```

server/cluster/topology_manager.go
```diff
@@ -81,6 +85,7 @@ type TopologyManagerImpl struct {
 	shardNodesMapping map[storage.ShardID][]storage.ShardNode // ShardID -> nodes of the shard
 	nodeShardsMapping map[string][]storage.ShardNode          // nodeName -> shards of the NodeName
 	// ShardView in memory.
+	shardViews         []storage.ShardView
```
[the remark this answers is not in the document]
ZuLiangWang: Unlike L89, it will be directly returned to the outside, and it cannot be modified by the outside.

server/coordinator/node_picker.go
ZuLiangWang: That's right, I thought about it again. The more important reason is that the scheduler should be scheduled based on the static topology of the current version instead of dynamically obtaining the latest metadata. Maybe `ShardPicker` also needs similar modifications.

server/coordinator/scheduler/scheduler.go
ZuLiangWang: Struct name could not start with package name, it will cause golint failed, just rename `scheduleReason` to `Reason`.

server/coordinator/scheduler/scheduler.go
ZuLiangWang: The function comment must start with function name, so it is `Schedule` rather than `Scheduler`...

server/coordinator/node_picker.go
ZuLiangWang: The above implementation will always return the same node in some cases, I tweaked a bit to fix this.

server/cluster/topology_manager.go
```diff
@@ -451,6 +461,13 @@ func (m *TopologyManagerImpl) CreateShardViews(ctx context.Context, createShardV
 	return nil
 }
 
+func (m *TopologyManagerImpl) GetShardViews() []storage.ShardView {
```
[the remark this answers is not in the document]
ZuLiangWang: The current implementation returns a copy of the value already, so it doesn't need to be modified.
This function is not `GetShardViewsByID`, and the purpose of this function is to obtain all shardViews. If the parameter a is added, the corresponding id must be obtained before calling, which seems unreasonable.

server/coordinator/node_picker.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (p *RandomNodePicker) PickNode(_ context.Context, registerNodes []cluster.RegisteredNode) (cluster.RegisteredNode, error) {
+	// Filter invalid nodes.
+	numOnlineNodes := int64(0)
+	for idx := range registerNodes {
+		if registerNodes[idx].IsOnline() {
+			numOnlineNodes++
+		}
+	}
+	if numOnlineNodes < 0 {
```
ShiKaiWi, line 32: It can't be negative, do you mean:
```suggestion
	if numOnlineNodes == 0 {
```

server/coordinator/node_picker.go
```diff
@@ -0,0 +1,56 @@
...
+
+		if registerNodes[selectedIdx].IsOnline() {
+			return registerNodes[selectedIdx], nil
+		}
+		selectedIdx++
+
+		if selectedIdx >= int64(len(registerNodes)) {
+			selectedIdx = 0
+		}
+	}
```
ShiKaiWi, line 55:
```suggestion
  selectedOnlineIdx := int(randSelectedIdx.Int64())
  currOnlineIdx := -1
  for idx:=0; idx<len(registerNodes); idx++ {
    if currOnlineIdx == selectedIdx {
      return registeredNodes[idx], nil
    }

    if registerNodes[idx].IsOnline() {
      currOnlineIdx ++
    }
  }
```

server/coordinator/node_picker.go
```diff
@@ -0,0 +1,56 @@
...
+	PickNode(ctx context.Context, registerNodes []cluster.RegisteredNode) (cluster.RegisteredNode, error)
+}
+
+type RandomNodePicker struct{}
+
+func NewRandomNodePicker() NodePicker {
+	return &RandomNodePicker{}
+}
+
+func (p *RandomNodePicker) PickNode(_ context.Context, registerNodes []cluster.RegisteredNode) (cluster.RegisteredNode, error) {
```
ShiKaiWi, line 24:
```suggestion
func (p *RandomNodePicker) PickNode(_ context.Context, registeredNodes []cluster.RegisteredNode) (cluster.RegisteredNode, error) {
```

server/coordinator/scheduler/scheduler.go
ShiKaiWi: I guess the package name is `scheduler`, which is different from `schedule`.

server/cluster/topology_manager.go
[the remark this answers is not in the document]
ShiKaiWi: However, the data in the slice is still shared and mutable.

server/coordinator/scheduler/scheduler.go
ShiKaiWi: And the procedure should be made public? Otherwise, the result's procedure can't be accessible to caller from the other packages.

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -0,0 +1,146 @@
...
+		return err
+	}
+	ctxWithCancel, cancel := context.WithCancel(ctx)
+	for _, c := range clusters {
+		c := c
+		go func() {
+			for {
+				select {
+				case <-ctxWithCancel.Done():
+					return
```
ShiKaiWi, line 96: Shall we add some logs before exit the schedule loop?

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+}
+
+func (m *ManagerImpl) Stop(_ context.Context) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	if m.isRunning {
+		m.cancel()
+		m.registerSchedulers = []Scheduler{}
```
ShiKaiWi, line 68: Why do we have to clear the registered schedulers?

server/coordinator/scheduler/assign_shard_scheduler.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+		return Result{
+			p:      p,
+			Reason: AssignReason,
+		}, nil
+	}
+	return Result{}, nil
+}
+
+func contains(shardID storage.ShardID, shardNodes []storage.ShardNode) (bool, storage.ShardNode) {
```
ShiKaiWi, line 64:
```suggestion
func findNodeByShard(shardID storage.ShardID, shardNodes []storage.ShardNode) (bool, storage.ShardNode) {
```

server/coordinator/scheduler/assign_shard_scheduler.go
```diff
@@ -0,0 +1,71 @@
...
+			NewLeaderNodeName: newLeaderNode.Node.Name,
+			ClusterVersion:    topology.ClusterView.Version,
+		})
+		if err != nil {
+			return Result{}, err
+		}
+		return Result{
+			p:      p,
+			Reason: AssignReason,
+		}, nil
```
ShiKaiWi, line 59:
```suggestion
		if p, err := a.factory.CreateTransferLeaderProcedure(ctx, coordinator.TransferLeaderRequest{
			ClusterName:       topology.ClusterView.ClusterName,
			ShardID:           shardView.ShardID,
			OldLeaderNodeName: "",
			NewLeaderNodeName: newLeaderNode.Node.Name,
			ClusterVersion:    topology.ClusterView.Version,
		}); err != nil {
			return Result{}, err
		} else {
		        return Result{
			p:      p,
			Reason: AssignReason,
			}, nil
		}
```

server/coordinator/scheduler/assign_shard_scheduler.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (a AssignShardScheduler) Schedule(ctx context.Context, topology cluster.Topology) (Result, error) {
+	if topology.ClusterView.State != storage.ClusterStateStable {
+		return Result{}, cluster.ErrClusterStateInvalid
+	}
+
+	// Check if there is a shard without node mapping.
+	for _, shardView := range topology.ShardViews {
+		exists, _ := contains(shardView.ShardID, topology.ClusterView.ShardNodes)
```
ShiKaiWi, line 37: Maybe we can avoid copy caused by the loop:
```suggestion
	for idx := range topology.ShardViews {
	        shardID := topology.ShardViews[idx].ShardID
		exists, _ := contains(shardView.ShardID, topology.ClusterView.ShardNodes)
```

server/coordinator/scheduler/scheduler_manager.go
ZuLiangWang: `schedulerManager` will start after elected and stop before transfer,  if we do not clear the registered schedulers, it will recall `initRegister` in start when this server is elected to leader again, which may be add duplicate schedulers.

server/coordinator/node_picker.go
ZuLiangWang: Reimplemented it as suggested.

server/coordinator/scheduler/assign_shard_scheduler.go
ZuLiangWang: This violates the rules of golint.
https://zchee.github.io/golang-wiki/CodeReviewComments/#indent-error-flow

server/cluster/cluster.go
```diff
@@ -438,13 +438,25 @@ func (c *Cluster) GetClusterState() storage.ClusterState {
...
+}
+
 func (c *Cluster) UpdateClusterView(ctx context.Context, state storage.ClusterState, shardNodes []storage.ShardNode) error {
 	if err := c.topologyManager.UpdateClusterView(ctx, state, shardNodes); err != nil {
 		return errors.WithMessage(err, "update cluster view")
 	}
 	return nil
 }
 
+func (c *Cluster) GetShardViews() []storage.ShardView {
```
ShiKaiWi, line 452: Why not name it as `RangeShardViews` too?

go.mod
```diff
@@ -101,3 +101,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+replace (
```
ShiKaiWi, line 105: Maybe it should be updated before merging this PR.

server/coordinator/node_picker.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if onlineNodeLength == 0 {
+		return cluster.RegisteredNode{}, errors.WithMessage(ErrNodeNumberNotEnough, "online node length must bigger than 0")
+	}
+
+	randSelectedIdx, err := rand.Int(rand.Reader, big.NewInt(int64(onlineNodeLength)))
+	if err != nil {
+		return cluster.RegisteredNode{}, errors.WithMessage(err, "generate random node index")
+	}
+	selectIdx := randSelectedIdx.Int64()
```
ShiKaiWi, line 41: Maybe you should convert the `selectIdx` to `int` rather than converting other variables' types.

server/coordinator/node_picker_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/CeresDB/ceresmeta/server/cluster"
+	"github.com/CeresDB/ceresmeta/server/storage"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	NodeLength = 3
+)
+
+func TestRandomNodePicker(t *testing.T) {
```
ShiKaiWi, line 19: Add some corner cases to test, e.g.:
- empty source nodes;
- non-empty source nodes with no online nodes in them;

server/coordinator/scheduler/assign_shard_scheduler.go
```diff
@@ -0,0 +1,73 @@
...
+		return ScheduleResult{
+			p:      p,
+			Reason: AssignReason,
+		}, nil
+
+	}
+	return ScheduleResult{}, nil
+}
+
+func findNodeByShard(shardID storage.ShardID, shardNodes []storage.ShardNode) (bool, storage.ShardNode) {
```
ShiKaiWi, line 66: The returned values may follow the convention of the golang:
```suggestion
func findNodeByShard(shardID storage.ShardID, shardNodes []storage.ShardNode) (storage.ShardNode, bool) {
```

server/coordinator/scheduler/scheduler.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"context"
+
+	"github.com/CeresDB/ceresmeta/server/cluster"
+	"github.com/CeresDB/ceresmeta/server/coordinator/procedure"
+)
+
+type ScheduleResult struct {
+	p procedure.Procedure
```
ShiKaiWi, line 13: If the `p` is not public, how to use it outside this package?

server/coordinator/scheduler/scheduler_manager.go
ShiKaiWi: If so, I guess it is better to clear the slice by:
```suggestion
		m.registerSchedulers = m.registerSchedulers[:0]
```

go.mod
ZuLiangWang: Here is the corresponding proto pull request, PTAL https://github.com/CeresDB/ceresdbproto/pull/71

ZuLiangWang: I re-thought the current implementation. Procedure should no longer hold a reference to Cluster, so there is no need to pass in ClusterName here. The information of the entire cluster is directly passed in through Request. This change is not included in this pull request.

server/cluster/cluster.go
ZuLiangWang: Remove it, and add `GetTopology` in `GetClusterSnapshot`.

ShiKaiWi (approved):
LGTM

PR #149, opened 2023-03-23
refactor: implement entry lock

ZuLiangWang:

# Rationale for this change
 In order to achieve concurrency of procedures at the shard level, we need to implement locks on the shard entity for procedures.

* Implement entry lock in `entry_lock.go`, which support try lock in multi shards.
* Add unit test for entry lock.

None.

# How does this change test
Pass the newly created unit test.

### DIFF
--- a/server/coordinator/lock/entry_lock.go
+++ b/server/coordinator/lock/entry_lock.go
@@ -1,0 +1,54 @@
+// Copyright 2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+package lock
+
+import (
+	"fmt"
+	"sync"
+)
+
+type EntryLock struct {
+	lock       sync.Mutex
+	entryLocks map[uint64]struct{}
+}
+
+func NewEntryLock(initCapacity int) EntryLock {
+	return EntryLock{
+		entryLocks: make(map[uint64]struct{}, initCapacity),
+	}
+}
+
+func (l *EntryLock) TryLock(locks []uint64) bool {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	for _, lock := range locks {
+		_, exists := l.entryLocks[lock]
+		if exists {
+			return false
+		}
+	}
+
+	for _, lock := range locks {
+		l.entryLocks[lock] = struct{}{}
+	}
+
+	return true
+}
+
+func (l *EntryLock) UnLock(locks []uint64) {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	for _, lock := range locks {
+		_, exists := l.entryLocks[lock]
+		if !exists {
+			panic(fmt.Sprintf("try to unlock nonexistent lock, exists locks:%v, unlock locks:%v", l.entryLocks, locks))
+		}
+	}
+
+	for _, lock := range locks {
+		delete(l.entryLocks, lock)
+	}
+
+}
--- a/server/coordinator/lock/entry_lock_test.go
+++ b/server/coordinator/lock/entry_lock_test.go
@@ -1,0 +1,42 @@
+// Copyright 2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+package lock
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


server/coordinator/entry_lock.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+func (l *EntryLockImpl) TryLock(locks []uint64) bool {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	sort.Slice(locks, func(i, j int) bool {
+		return i > j
+	})
```
ShiKaiWi, line 30: No need to do sort because all the locks are acquired atomically.

server/coordinator/entry_lock.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type EntryLock interface {
+	TryLock(locks []uint64) bool
+	ReleaseLock(locks []uint64) bool
+}
+
+type EntryLockImpl struct {
+	lock       sync.Mutex
+	entryLocks map[uint64]bool
```
ShiKaiWi, line 15: Actually, we can use empty struct `struct{}` as the value.

server/coordinator/entry_lock.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func NewEntryLock() EntryLock {
```
ShiKaiWi, line 18: Capacity for initialization may help reduce reallocation when being used.

server/coordinator/entry_lock.go
```diff
@@ -0,0 +1,62 @@
+package coordinator
+
+import (
+	"sort"
+	"sync"
+)
+
+type EntryLock interface {
+	TryLock(locks []uint64) bool
+	ReleaseLock(locks []uint64) bool
```
ShiKaiWi, line 10: In `golang` style, `Unlock` seems a better name:
```suggestion
	Unlock(locks []uint64) bool
```
BTW, I guess no need to abstract an interface for it.

server/coordinator/entry_lock.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (l *EntryLockImpl) ReleaseLock(locks []uint64) bool {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	for _, lock := range locks {
+		_, exists := l.entryLocks[lock]
+		if !exists {
+			return false
+		}
```
ShiKaiWi, line 54: A better behavior is just panic if the caller unlocks a wrong batch, and no need to return a boolean to tell the caller whether to succeed in unlocking.

server/coordinator/entry_lock.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (l *EntryLock) UnLock(locks []uint64) {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	for _, lock := range locks {
+		_, exists := l.entryLocks[lock]
+		if !exists {
+			panic(fmt.Sprintf("try to unlock nonexistent lock, exists locks: %v, unlock locks: %v", l.entryLocks, locks))
```
ShiKaiWi:
```suggestion
			panic(fmt.Sprintf("try to unlock nonexistent lock, exists locks:%v, unlock locks:%v", l.entryLocks, locks))
```

server/coordinator/entry_lock_test.go
```diff
@@ -0,0 +1,36 @@
...
+	lock3 := []uint64{3, 4, 5}
+	result = lock.TryLock(lock2)
+	re.Equal(true, result)
+	result = lock.TryLock(lock2)
+	re.Equal(false, result)
+	result = lock.TryLock(lock3)
+	re.Equal(false, result)
+	lock.UnLock(lock2)
+	result = lock.TryLock(lock2)
+	re.Equal(true, result)
```
ShiKaiWi: Add test for wrong unlock.

ShiKaiWi (approved):
LGTM

PR #150, opened 2023-03-24
refactor: implement delay queue

ZuLiangWang:

# Rationale for this change
 In order to achieve concurrency of procedures at the shard level, we need to support submit waiting procedures, in order to manage these waiting procedures, support for creating a delay queue is required.

* Add `DelayQueue` in `delay_queue.go`.
* Add unit test for `DelayQueue`.

None.

# How does this change test
Pass the newly created unit test.


server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,73 @@
...
+	defer q.lock.Unlock()
+
+	if len(q.procedures) >= q.maxLen {
+		return errors.WithMessage(ErrQueueFull, fmt.Sprintf("queue max length is %d", q.maxLen))
+	}
+
+	for _, procedureWithTime := range q.procedures {
+		if procedureWithTime.procedure.ID() == p.ID() {
+			return errors.WithMessage(ErrOfferDuplicatedProcedure, fmt.Sprintf("procedure has been offered, %v", p))
+		}
```
ShiKaiWi, line 50: If it is required to support duplicate check, how about add a hashmap for it?

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/CeresDB/ceresmeta/server/coordinator/procedure"
+	"github.com/pkg/errors"
+)
+
+type ProcedureWithTime struct {
```
ShiKaiWi, line 14: How about `ProcedureScheduleEntry`?

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,73 @@
...
+	procedure procedure.Procedure
+	time      time.Time
```
ShiKaiWi, line 16: How about `run_after`?

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (q *ProcedureDelayQueue) Len() int {
+	q.lock.RLock()
+	defer q.lock.RUnlock()
+
+	return len(q.procedures)
+}
+
+func (q *ProcedureDelayQueue) Offer(p procedure.Procedure, delaySec time.Duration) error {
```
ShiKaiWi, line 39: It is rare to name the methods of a Queue as Offer and Poll. How about Push and Pop?
```suggestion
func (q *ProcedureDelayQueue) Push(p procedure.Procedure, delay time.Duration) error {
```

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (q *ProcedureDelayQueue) Poll() procedure.Procedure {
+	q.lock.Lock()
+	defer q.lock.Unlock()
+
+	for i, procedureWithTime := range q.procedures {
+		if procedureWithTime.time.Unix() <= time.Now().Unix() {
+			procedure := procedureWithTime.procedure
+			q.procedures = append(q.procedures[:i], q.procedures[i+1:]...)
```
ShiKaiWi, line 68: Such implementation is not very efficient. And maybe we should use [`Heap`](https://pkg.go.dev/container/heap) instead.

server/coordinator/queue/delay_queue.go
ZuLiangWang: Added a separate map to maintain this mapping relationship.

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (q *HeapPriorityQueue) Len() int {
+	return len(q.procedures)
+}
+
+func (q *HeapPriorityQueue) Less(i, j int) bool {
+	if q.procedures[i].runAfter.Before(q.procedures[j].runAfter) {
+		return true
+	}
```
ShiKaiWi, line 40: I guess it should the negative value to make the min runAfter lie at the end of the `q.procedures`:
```suggestion
	if q.procedures[j].runAfter.Before(q.procedures[i].runAfter) {
		return true
	}
```

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if q.heapQueue.Len() == 0 {
+		return nil
+	}
+
+	entry := heap.Pop(q.heapQueue).(*ProcedureScheduleEntry)
+	if time.Now().Before(entry.runAfter) {
+		heap.Push(q.heapQueue, entry)
+		return nil
+	}
```
ShiKaiWi, line 113: Maybe we can add a method called `Peek` on the `HeapPriorityQueue`, with which we can do time check without mutating the heap.

server/coordinator/queue/delay_queue_test.go
```diff
@@ -0,0 +1,81 @@
...
+	queue := NewProcedureDelayQueue(3)
+	err := queue.Push(testProcedure0, time.Second)
+	re.NoError(err)
+	err = queue.Push(testProcedure0, time.Second)
+	re.Error(err)
+	err = queue.Push(testProcedure1, time.Second)
+	re.NoError(err)
+	err = queue.Push(testProcedure2, time.Second)
+	re.NoError(err)
+	err = queue.Push(testProcedure3, time.Second)
```
ShiKaiWi, line 53: Maybe we should test these procedures with different delay?
BTW, use millisecond instead of second to make this test run faster.

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,117 @@
...
+		if procedureWithTime.procedure.ID() == p.ID() {
+			return errors.WithMessage(ErrPushDuplicatedProcedure, fmt.Sprintf("procedure has been pushed, %v", p))
+		}
+	}
+
+	heap.Push(q.heapQueue, &ProcedureScheduleEntry{
+		procedure: p,
+		runAfter:  time.Now().Add(delay),
+	})
+	q.index[p.ID()] = p
```
ShiKaiWi, line 96: It seems this `index` is not used for duplicate checking?

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,117 @@
...
+
+type ProcedureScheduleEntry struct {
+	procedure procedure.Procedure
+	runAfter  time.Time
+}
+
+type ProcedureDelayQueue struct {
+	lock      sync.RWMutex
+	heapQueue *HeapPriorityQueue
+	index     map[uint64]procedure.Procedure
```
ShiKaiWi, line 23: It looks better if put it as:
```suggestion
	existingProcs     map[uint64]struct{}
```

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,117 @@
...
+	maxLen    int
+}
+
+type HeapPriorityQueue struct {
```
ShiKaiWi, line 27: No need to export it outside.

ShiKaiWi (changes requested):

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,117 @@
...
+	"container/heap"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/CeresDB/ceresmeta/server/coordinator/procedure"
+	"github.com/pkg/errors"
+)
+
+type ProcedureScheduleEntry struct {
```
ShiKaiWi, line 15: No need to export this struct.

ShiKaiWi (commented):
@ZuLiangWang For the key structure, some comments may help others and yourself in the future to understand the code. Hope for more comments in this PR and future PRs.

server/coordinator/queue/delay_queue.go
ZuLiangWang: This index is indeed used to detect duplicate data, I missed it when modifying, and now it is fixed.

server/coordinator/queue/delay_queue.go
ZuLiangWang: The dequeue order of elements is determined by the less method. If `procedures[i] < procedures[j]`, the `smallest` element will be dequeued first. If `procedures[i] > procedures[j]`, the `largest` element will be dequeued first. So the current the order is correct.

ZuLiangWang: This order is difficult to understand, I will add some notes separately.

server/coordinator/queue/delay_queue.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (q *heapPriorityQueue) Len() int {
+	return len(q.procedures)
+}
+
+// The dequeue order of elements is determined by the less method.
+// When return procedures[i].runAfter < procedures[j].runAfter, the element with smallest will be pop first.
+func (q *heapPriorityQueue) Less(i, j int) bool {
+	return q.procedures[i].runAfter.Before(q.procedures[j].runAfter)
```
ShiKaiWi, line 44: In order to make the smallest element located at the last in the slice, the compare must be negative, that is to say:
```
```suggestion
	return q.procedures[j].runAfter.Before(q.procedures[i].runAfter)
```
```

**Review by ShiKaiWi — CHANGES_REQUESTED** [review]

File: server/coordinator/queue/delay_queue.go (hunk 13 above)
Comment (ShiKaiWi) [reply server/coordinator/queue/delay_queue.go:44]: This is my fault. This PR is ok now.

File: server/coordinator/queue/delay_queue.go [hunk 14]
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (q *heapPriorityQueue) Peek() any {
+	length := len(q.procedures)
+	if length == 0 {
+		return nil
+	}
+	item := q.procedures[0]
+	return item
+}
```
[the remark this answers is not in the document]
ShiKaiWi: Sorry, It's my fault.

ShiKaiWi (approved):

PR #154, opened 2023-04-12
refactor: refactor shard picker

ZuLiangWang:

Closes #

# Rationale for this change
The current shard picker uses the cluster instance to dynamically obtain the latest cluster topology. This is wrong. It should be scheduled based on the snapshot of the cluster. In this pull request, use snapshot to replace cluster.

* Modify the `ShardPicker` interface and add the `Snapshot` parameter.
* Modify the implementation to adapt to the current interface.

None.

# How does this change test
Unit tests are currently not working properly and will be fixed in a subsequent pull request.

### DIFF
--- a/server/coordinator/shard_picker.go
+++ b/server/coordinator/shard_picker.go
@@ -44,7 +44,7 @@ func (p *RandomBalancedShardPicker) PickShards(_ context.Context, snapshot metad
 
 	if !enableDuplicateNode {
 		if len(nodeShardsMapping) < expectShardNum {
-			return []storage.ShardNode{}, errors.WithMessagef(ErrNodeNumberNotEnough, "number of nodes is:%d, expecet number of shards is:%d", len(nodeShardsMapping), expectShardNum)
+			return nil, errors.WithMessagef(ErrNodeNumberNotEnough, "number of nodes is:%d, expecet number of shards is:%d", len(nodeShardsMapping), expectShardNum)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


server/coordinator/shard_picker.go
```diff
@@ -18,59 +18,52 @@ import (
...
+			nodeShardsMapping[shardNode.NodeName] = shardNodes
 		}
-		nodeShardsMapping[nodeShard.ShardNode.NodeName] = append(nodeShardsMapping[nodeShard.ShardNode.NodeName], nodeShard)
+		nodeShardsMapping[shardNode.NodeName] = append(nodeShardsMapping[shardNode.NodeName], shardNode)
 	}
 
 	if !enableDuplicateNode {
 		if len(nodeShardsMapping) < expectShardNum {
-			return []metadata.ShardNodeWithVersion{}, errors.WithMessagef(ErrNodeNumberNotEnough, "number of nodes is:%d, expecet number of shards is:%d", len(nodeShardsMapping), expectShardNum)
+			return []storage.ShardNode{}, errors.WithMessagef(ErrNodeNumberNotEnough, "number of nodes is:%d, expecet number of shards is:%d", len(nodeShardsMapping), expectShardNum)
```
ShiKaiWi, line 47:
```suggestion
			return nil, errors.WithMessagef(ErrNodeNumberNotEnough, "number of nodes is:%d, expecet number of shards is:%d", len(nodeShardsMapping), expectShardNum)
```

server/coordinator/shard_picker.go
```diff
@@ -81,7 +74,7 @@ func (p *RandomBalancedShardPicker) PickShards(ctx context.Context, clusterName
 
 			selectNodeIndex, err := rand.Int(rand.Reader, big.NewInt(int64(len(nodeNames))))
 			if err != nil {
-				return []metadata.ShardNodeWithVersion{}, errors.WithMessage(err, "generate random node index")
+				return []storage.ShardNode{}, errors.WithMessage(err, "generate random node index")
```
ShiKaiWi, line 77:
```suggestion
				return nil, errors.WithMessage(err, "generate random node index")
```

PR #156, opened 2023-04-17
refactor: refactor unit tests

ZuLiangWang:

Closes #

# Rationale for this change
Connect all the pull requests and modifications that were merged before, so that the service can run normally, and solve the problems of compilation, unit testing and CI.

* Concatenate all previous modifications to solve compilation errors.
* Adapt unit tests to the current version of the code so that all unit tests run and pass.
* Solve some errors detected by golint.

None.

# How does this change test
Pass all test units.

### DIFF
--- a/server/cluster/metadata/cluster_metadata.go
+++ b/server/cluster/metadata/cluster_metadata.go
@@ -273,8 +273,33 @@ func (c *ClusterMetadata) CreateTableMetadata(ctx context.Context, request Creat
 		Table: table,
 	}
 
-	log.Info("create table metadata succeed", zap.String("cluster", c.Name()), zap.String("result", fmt.Sprintf("%+v", res)), zap.Object("result", res))
+	log.Info("create table metadata succeed", zap.String("cluster", c.Name()), zap.String("result", fmt.Sprintf("%+v", res)))
 	return res, nil
+}
+
+func (c *ClusterMetadata) DropTableMetadata(ctx context.Context, schemaName, tableName string) (DropTableMetadataResult, error) {
+	log.Info("drop table start", zap.String("cluster", c.Name()), zap.String("schemaName", schemaName), zap.String("tableName", tableName))
+
+	table, ok, err := c.tableManager.GetTable(schemaName, tableName)
+	if err != nil {
+		return DropTableMetadataResult{}, errors.WithMessage(err, "get table")
+	}
+
+	if !ok {
+		return DropTableMetadataResult{}, ErrTableNotFound
+	}
+
+	err = c.tableManager.DropTable(ctx, schemaName, tableName)
+	if err != nil {
+		return DropTableMetadataResult{}, errors.WithMessage(err, "table manager drop table")
+	}
+
+	ret := DropTableMetadataResult{
+		table,
+	}
+	log.Info("drop table metadata success", zap.String("cluster", c.Name()), zap.String("schemaName", schemaName), zap.String("tableName", tableName), zap.String("result", fmt.Sprintf("%+v", ret)))
+
+	return ret, nil
 }
 
 func (c *ClusterMetadata) CreateTable(ctx context.Context, request CreateTableRequest) (CreateTableResult, error) {
@@ -341,7 +366,7 @@ func (c *ClusterMetadata) RegisterNode(ctx context.Context, registeredNode Regis
 	}
 
 	// Update shard node mapping.
-	var shardNodes map[string][]storage.ShardNode
+	shardNodes := make(map[string][]storage.ShardNode, len(registeredNode.ShardInfos))
 	for _, shardInfo := range registeredNode.ShardInfos {
 		shardNodes[registeredNode.Node.Name] = append(shardNodes[registeredNode.Node.Name], storage.ShardNode{
 			ID:        shardInfo.ID,
@@ -349,7 +374,7 @@ func (c *ClusterMetadata) RegisterNode(ctx context.Context, registeredNode Regis
 			NodeName:  registeredNode.Node.Name,
 		})
 	}
-	if err := c.updateClusterViewByNode(ctx, shardNodes); err != nil {
+	if err := c.UpdateClusterViewByNode(ctx, shardNodes); err != nil {
 		return errors.WithMessage(err, "update cluster view failed")
 	}
 
@@ -426,10 +451,11 @@ func (c *ClusterMetadata) RouteTables(_ context.Context, schemaName string, tabl
 		table := tables[tableID]
 		routeEntries[table.Name] = RouteEntry{
 			Table: TableInfo{
-				ID:         table.ID,
-				Name:       table.Name,
-				SchemaID:   table.SchemaID,
-				SchemaName: schemaName,
+				ID:            table.ID,
+				Name:          table.Name,
+				SchemaID:      table.SchemaID,
+				SchemaName:    schemaName,
+				PartitionInfo: table.PartitionInfo,
 			},
 			NodeShards: nodeShardsResult,
 		}
@@ -501,9 +527,16 @@ func (c *ClusterMetadata) UpdateClusterView(ctx context.Context, state storage.C
 	return nil
 }
 
-func (c *ClusterMetadata) updateClusterViewByNode(ctx context.Context, shardNodes map[string][]storage.ShardNode) error {
+func (c *ClusterMetadata) UpdateClusterViewByNode(ctx context.Context, shardNodes map[string][]storage.ShardNode) error {
 	if err := c.topologyManager.UpdateClusterViewByNode(ctx, shardNodes); err != nil {
 		return errors.WithMessage(err, "update cluster view")
+	}
+	return nil
+}
+
+func (c *ClusterMetadata) DropShardNode(ctx context.Context, shardNodes []storage.ShardNode) error {
+	if err := c.topologyManager.DropShardNodes(ctx, shardNodes); err != nil {
+		return errors.WithMessage(err, "drop shard nodes")
 	}
 	return nil
 }
@@ -531,6 +564,21 @@ func (c *ClusterMetadata) GetClusterSnapshot() Snapshot {
 func (c *ClusterMetadata) Init(ctx context.Context) error {
 	c.lock.Lock()
 	defer c.lock.Unlock()
+
+	var createShardViews []CreateShardView
+	for i := uint32(0); i < c.metaData.ShardTotal; i++ {
+		shardID, err := c.AllocShardID(ctx)
+		if err != nil {
+			return errors.WithMessage(err, "alloc shard id failed")
+		}
+		createShardViews = append(createShardViews, CreateShardView{
+			ShardID: storage.ShardID(shardID),
+			Tables:  []storage.TableID{},
+		})
+	}
+	if err := c.topologyManager.CreateShardViews(ctx, createShardViews); err != nil {
+		return errors.WithMessage(err, "create shard view")
+	}
 
 	return c.topologyManager.InitClusterView(ctx)
 }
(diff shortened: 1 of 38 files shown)


server/cluster/metadata/cluster_metadata.go
```diff
@@ -532,6 +557,21 @@ func (c *ClusterMetadata) Init(ctx context.Context) error {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	var createShardViews []CreateShardView
```
ShiKaiWi, line 560: Shall we preallocate the capacity for it?

server/cluster/metadata/table_manager_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2022 CeresDB Project Authors. Licensed under Apache-2.0.
 
 // nolint
-package metadata
+package metadata_test
```
ShiKaiWi, line 4: Is there this style in the Golang?

server/cluster/metadata/topology_manager.go
```diff
@@ -261,6 +270,14 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
 		}
 	}
 
+	for i, oldShardView := range m.shardViews {
+		if oldShardView.ShardID == shardID {
+			m.shardViews[i] = m.shardViews[len(m.shardViews)-1]
+			m.shardViews = append(m.shardViews, newShardView)
+			break
+		}
+	}
```
ShiKaiWi, line 279: Avoid clone the shard view when looping it:
```suggestion
	for i := range m.shardViews {
		if m.shardViews[i].ShardID == shardID {
			m.shardViews[i] = newShardView
			break
		}
	}
```

And how to deal with case where the shard is not found in the shardViews?

BTW, maybe we can define it as a method for reuse here and L206.

server/cluster/metadata/types.go
```diff
@@ -74,6 +74,10 @@ type DropTableResult struct {
 	ShardVersionUpdate []ShardVersionUpdate
 }
 
+type DropTableMetadataResult struct {
+	Table storage.Table
```
ShiKaiWi, line 78: When to use this field?

server/coordinator/factory.go
```diff
@@ -149,39 +140,44 @@ func (f *Factory) makeCreatePartitionTableProcedure(ctx context.Context, request
...
 	}
 
-	getNodeShardResult, err := request.Cluster.GetNodeShards(ctx)
-	if err != nil {
-		log.Error("cluster get node shard result")
-		return nil, err
-	}
+	snapshot := request.ClusterMetadata.GetClusterSnapshot()
 
 	nodeNames := make(map[string]int)
```
ShiKaiWi, line 145: Preallocate memory for it.

server/coordinator/factory.go
```diff
@@ -149,39 +140,44 @@ func (f *Factory) makeCreatePartitionTableProcedure(ctx context.Context, request
...
+	subTableShards, err := f.shardPicker.PickShards(ctx, snapshot, len(request.SourceReq.PartitionTableInfo.SubTableNames), true)
 	if err != nil {
-		return nil, errors.WithMessage(err, "pick partition table shards")
+		return nil, errors.WithMessage(err, "pick sub table shards")
 	}
 
-	subTableShards, err := f.shardPicker.PickShards(ctx, request.Cluster.Name(), len(request.SourceReq.PartitionTableInfo.SubTableNames), true)
-	if err != nil {
-		return nil, errors.WithMessage(err, "pick data table shards")
+	shardNodeWithVersion := make([]metadata.ShardNodeWithVersion, 0, len(subTableShards))
```
ShiKaiWi, line 155:
```suggestion
	shardNodesWithVersion := make([]metadata.ShardNodeWithVersion, 0, len(subTableShards))
```

server/coordinator/factory.go
```diff
@@ -149,39 +140,44 @@ func (f *Factory) makeCreatePartitionTableProcedure(ctx context.Context, request
...
+	for _, subTableShard := range subTableShards {
+		for _, shardView := range snapshot.Topology.ShardViews {
```
ShiKaiWi, line 157: The `ShardView` seems always be used to find shard view by shard id, is it possible to declare it as a map?

server/cluster/metadata/table_manager_test.go
ZuLiangWang: Yes, this is a little trick for writing unit tests in Golang. Test files that declare a package with the suffix "_test" will be compiled as a separate package, and then linked and run with the main test binary.
This has two advantages. First, different package names cannot directly refer to internal variables and functions, which enables black-box testing. Second, it avoids circular dependencies.
More information can be found in this issue: https://github.com/golang/go/issues/25223

server/cluster/metadata/types.go
ZuLiangWang: It is used to tell the user the information of the table being dropped when the partition table is dropped.

server/coordinator/factory.go
ZuLiangWang: Modify it to a map, and initialize the map when it is created.

server/cluster/metadata/table_manager_test.go
ShiKaiWi: Thanks! Learned a lot from it. However, how to test the private methods if so?

server/cluster/metadata/topology_manager.go
```diff
@@ -231,14 +248,15 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
 	}
 
 	// Update shardView in storage.
+	newShardView := storage.ShardView{
+		ShardID:   shardView.ShardID,
+		Version:   prevVersion + 1,
+		TableIDs:  newTableIDs,
+		CreatedAt: uint64(time.Now().UnixMilli()),
+	}
```
ShiKaiWi, line 256: How about declaring a new method for ShardView to avoid calling `time.Now().UnixMilli()` again and again?

server/cluster/metadata/topology_manager.go
```diff
@@ -390,6 +410,30 @@ func (m *TopologyManagerImpl) GetShardNodes() GetShardNodesResult {
...
+}
+
+func contains(shardNodes []storage.ShardNode, originShardNode storage.ShardNode) bool {
+	for _, dropShardNode := range shardNodes {
+		if originShardNode.NodeName == dropShardNode.NodeName && originShardNode.ID == dropShardNode.ID {
+			return true
+		}
+	}
+	return false
+}
```
ShiKaiWi, line 435: When you use slice of a struct rather than a pointer, you have to be careful to not introduce copy when looping over it.

server/coordinator/node_picker.go
```diff
@@ -6,11 +6,16 @@ import (
...
 	"crypto/rand"
 	"math/big"
+	"time"
 
 	"github.com/CeresDB/ceresmeta/server/cluster/metadata"
 	"github.com/pkg/errors"
 )
 
+const (
+	expiredThreshold = time.Second * 10
```
ShiKaiWi, line 16: I guess whether a node is expired should be decided in the `node_picker`. Maybe you should move this threshold to the metadata package.

server/coordinator/node_picker_test.go
```diff
@@ -29,7 +30,7 @@ func TestRandomNodePicker(t *testing.T) {
 
 	for i := 0; i < nodeLength; i++ {
 		nodes = append(nodes, metadata.RegisteredNode{
-			Node:       storage.Node{Name: strconv.Itoa(i), State: storage.NodeStateOffline},
+			Node:       storage.Node{Name: strconv.Itoa(i), LastTouchTime: uint64(time.Now().Unix() - int64(time.Minute))},
```
ShiKaiWi, line 33: Why not do `uint64(time.Now().Unix() - int64(time.Minute))` just once and use it everywhere?

server/coordinator/procedure/ddl/create_table_common_util.go
```diff
@@ -4,6 +4,9 @@ package ddl
 
 import (
 	"context"
+	"fmt"
+	"github.com/CeresDB/ceresmeta/pkg/log"
```
ShiKaiWi, line 8: Format the importing lines.

server/coordinator/procedure/ddl/create_table_common_util.go
```diff
@@ -36,6 +39,8 @@ func CreateTableMetadata(ctx context.Context, c *metadata.ClusterMetadata, schem
 }
 
 func CreateTableOnShard(ctx context.Context, c *metadata.ClusterMetadata, dispatch eventdispatch.Dispatch, shardID storage.ShardID, request eventdispatch.CreateTableOnShardRequest) error {
+	log.Info("createTableOnShard", zap.String("req", fmt.Sprintf("%v", request)))
+	log.Info("createTableOnShard", zap.Uint64("curVersion", request.UpdateShardInfo.CurrShardInfo.Version), zap.Uint64("preVersion", request.UpdateShardInfo.PrevVersion))
```
ShiKaiWi, line 43: Why not print them in the same log?

server/coordinator/procedure/ddl/droppartitiontable/drop_partition_table.go
```diff
@@ -59,23 +60,65 @@ type Procedure struct {
...
-		fsm:    fsm,
-		params: params,
+		fsm:                fsm,
+		params:             params,
+		relatedVersionInfo: relatedVersionInfo,
+	}, nil
+}
+
+func buildRelatedVersionInfo(params ProcedureParams) (procedure.RelatedVersionInfo, error) {
+	shardViewWithVersion := make(map[storage.ShardID]uint64, 0)
```
ShiKaiWi, line 90: Move this to above the L97 and preallocate memory for it.

server/coordinator/procedure/operation/transferleader/transfer_leader.go
```diff
@@ -106,22 +109,31 @@ func NewProcedure(params ProcedureParams) (procedure.Procedure, error) {
...
 }
 
 func validateClusterTopology(topology metadata.Topology, shardID storage.ShardID, oldLeaderNodeName string) error {
+	_, found := topology.ShardViewsMapping[shardID]
+	if !found {
+		log.Error("shard not found", zap.Uint64("shardID", uint64(shardID)))
+		return metadata.ErrShardNotFound
+	}
+	if len(oldLeaderNodeName) == 0 {
+		return nil
```
ShiKaiWi, line 135: Isn't this an error for transfer_leader procedure?

server/cluster/metadata/table_manager_test.go
ZuLiangWang: I do not want to test private method, I think unit testing only needs to ensure the correctness of the exposed functions. If the private function is huge or complex or important enough to require its own tests, just put it in another struct and make it public there.

server/coordinator/procedure/operation/transferleader/transfer_leader.go
ZuLiangWang: TransferLeader procedure ensure the OldLeaderNodeName is empty, which means that the shard has not been assigned to any node, which will appear when the cluster is initialized.

server/cluster/metadata/topology_manager.go
ZuLiangWang: Yes... maybe it should be written according to the principle similar to function visibility(Use private first in any case, unless the requirement cannot be met). In any case, use for index to traverse the slice first, unless it is found to avoid using a copy.
In the future, I will check all the slice traversal codes in the project individually, and fix them through a pull request.

server/cluster/metadata/table_manager_test.go
ShiKaiWi: I guess visibility should not be designed for tests, but let's leave the decision in the future and move on for now.

server/cluster/metadata/topology_manager.go
ShiKaiWi: Now, you have to call `generateCreateTime` again and again. Why not define a method called `NewShardView` in the storage package and generate the create timestamp there?

ShiKaiWi (approved):
LGTM

PR #160, opened 2023-04-25
fix: fix update cluster view when shardinfos is empty

ZuLiangWang:

Closes #

# Rationale for this change
When testing locally, it is found that shardNodes cannot be updated normally when the shardInfo in registerNode is empty.

1. Fix the bug that clusterView cannot be updated normally when the node is registered.

None.

# How does this change test
All unit tests pass, and local validation is correct.


ShiKaiWi (approved):
LGTM

PR #161, opened 2023-04-25
feat: renew the leader lease in a modern way

ShiKaiWi:

Closes #

# Rationale for this change
 The current way to renew leader lease is by a tick keep alive rpc, however, the etcd client supports a stream way to do such thing, which is more elegant and efficient.
 

- Renew the leader lease in a modern way;

None.

# How does this change test
Test locally. Before this pr, if the etcd log level is debug, then such logs will be printed again and again:
```
{"level":"debug","ts":"2023-04-25T16:22:53.891+0800","caller":"v3rpc/lease.go:118","msg":"failed to receive lease keepalive request from gRPC stream","error":"rpc error: code = Canceled desc = context canceled"}
```

### DIFF
--- a/server/member/lease.go
+++ b/server/member/lease.go
@@ -38,22 +38,6 @@ func newLease(rawLease clientv3.Lease, ttlSec int64) *lease {
 	}
 }
 
-type renewLeaseResult int
-
-const (
-	renewLeaseAlive renewLeaseResult = iota
-	renewLeaseFailed
-	renewLeaseExpired
-)
-
-func (r renewLeaseResult) failed() bool {
-	return r == renewLeaseFailed
-}
-
-func (r renewLeaseResult) alive() bool {
-	return r == renewLeaseAlive
-}
-
 func (l *lease) Grant(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctx, l.timeout)
 	defer cancel()
@@ -101,7 +85,7 @@ func (l *lease) KeepAlive(ctx context.Context) {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
 	go func() {
-		l.renewLeaseBg(ctx1, l.timeout/3, renewed)
+		l.renewLeaseBg(ctx1, renewed)
 		wg.Done()
 	}()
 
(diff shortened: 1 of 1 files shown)


server/member/lease.go
```diff
@@ -162,51 +162,51 @@ func (l *lease) getExpireTime() time.Time {
...
 // `renewLeaseBg` keeps the lease alive by periodically call `lease.KeepAliveOnce`.
 // The l.expireTime will be updated during renewing and the renew lease result (whether alive) will be told to caller by `renewed` channel.
-func (l *lease) renewLeaseBg(ctx context.Context, interval time.Duration, renewed chan<- bool) {
-	l.logger.Info("start renewing lease background", zap.Duration("interval", interval))
-	defer l.logger.Info("stop renewing lease background", zap.Duration("interval", interval))
+func (l *lease) renewLeaseBg(ctx context.Context, renewed chan<- bool) {
+	l.logger.Info("start renewing lease background", zap.Int64("lease-id", int64(l.ID)))
+	defer l.logger.Info("stop renewing lease background", zap.Int64("lease-id", int64(l.ID)))
+
+	keepaliveStream, err := l.rawLease.KeepAlive(ctx, l.ID)
```
ZuLiangWang, line 169:
```suggestion
	keepAliveStream, err := l.rawLease.KeepAlive(ctx, l.ID)
```

server/member/lease.go
```diff
@@ -101,7 +101,7 @@ func (l *lease) KeepAlive(ctx context.Context) {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
 	go func() {
-		l.renewLeaseBg(ctx1, l.timeout/3, renewed)
```
ZuLiangWang, line 104: `l.timeout` seems not be used now, maybe it should be removed from `lease`.

server/member/lease.go
```diff
@@ -162,51 +162,51 @@ func (l *lease) getExpireTime() time.Time {
...
-			if err != nil {
-				l.logger.Error("lease keep alive failed", zap.Error(err))
-				return renewLeaseFailed
-			}
-			if resp.TTL < 0 {
-				l.logger.Warn("lease is expired")
-				return renewLeaseExpired
+		start := time.Now()
+		select {
+		case resp, ok := <-keepaliveStream:
```
ZuLiangWang, line 178: What will happens if the `keepaAliveSteam` is disconnected?  In the previous implementation, if the retry succeeds before the lease timeout, a single `keepAliveOnce` failure will not lead to a new election, which has a certain degree of fault tolerance. But in the current implementation, once steam is disconnected, there will be no further attempts to renew the lease. Maybe we need try to reconnected when `keepaAliveSteam` is disconnected.

server/member/lease.go
```diff
@@ -101,7 +101,7 @@ func (l *lease) KeepAlive(ctx context.Context) {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
 	go func() {
-		l.renewLeaseBg(ctx1, l.timeout/3, renewed)
+		l.renewLeaseBg(ctx1, renewed)
```
ZuLiangWang, line 104:
```suggestion
	        // keepAliveInterval is set internally by `ETCD` client, we don't need to set it explicitly. 
                // nextKeepAlive := time.Now().Add((time.Duration(karesp.TTL) * time.Second) / 3.0)
		l.renewLeaseBg(ctx1, renewed)
```
Maybe we need add some comment to explain how the `keepAliveInterval` is added to `keepAliveSteam`.

server/member/lease.go
ShiKaiWi: Make senses. It should retry if the stream is broken.

ShiKaiWi:
Have no time to refactor this PR. close it for now.

PR #162, opened 2023-04-25
feat: add rebalanced shard scheduler

ZuLiangWang:

Closes #

# Rationale for this change
In the current failover implementation, if a node goes down, all its shards will be moved to other shards, and the shards will not be transferred back after the node is restarted, which will cause a lot of pressure on operation and maintenance. We need supports automatic load balancing.

1. Introduced the consistent hash implementation in the golang standard library, ref: https://github.com/golang/groupcache/blob/4a4ac3fbac33b83bb138f808c8945a2812023fc4/consistenthash/consistenthash.go
2. Implemented shard load balancing based on consistent hashing.

None.

# How does this change test
Pass all unit tests and local tests.

### DIFF
--- a/server/cluster/metadata/types.go
+++ b/server/cluster/metadata/types.go
@@ -135,8 +135,10 @@ func NewRegisteredNode(meta storage.Node, shardInfos []ShardInfo) RegisteredNode
 	}
 }
 
-func (n RegisteredNode) IsExpired(now int64) bool {
-	return now >= int64(expiredThreshold/time.Millisecond)+int64(n.Node.LastTouchTime)
+func (n RegisteredNode) IsExpired(now time.Time) bool {
+	expiredTime := time.UnixMilli(int64(n.Node.LastTouchTime)).Add(expiredThreshold)
+
+	return now.After(expiredTime)
 }
 
 func ConvertShardsInfoToPB(shard ShardInfo) *metaservicepb.ShardInfo {
--- a/server/cluster/metadata/types.go
+++ b/server/cluster/metadata/types.go
@@ -136,7 +136,7 @@ func NewRegisteredNode(meta storage.Node, shardInfos []ShardInfo) RegisteredNode
 }
 
 func (n RegisteredNode) IsExpired(now int64) bool {
-	return now >= int64(expiredThreshold)+int64(n.Node.LastTouchTime)
+	return now >= int64(expiredThreshold/time.Millisecond)+int64(n.Node.LastTouchTime)
 }
 
 func ConvertShardsInfoToPB(shard ShardInfo) *metaservicepb.ShardInfo {
--- a/server/coordinator/node_picker.go
+++ b/server/coordinator/node_picker.go
@@ -39,16 +39,16 @@ func (p *ConsistentHashNodePicker) PickNode(_ context.Context, shardID storage.S
 		}
 	}
 
-	if aliveNodeNumber == 0 {
+	if hashRing.IsEmpty() {
 		return metadata.RegisteredNode{}, errors.WithMessage(ErrNodeNumberNotEnough, "at least one online nodes is required")
 	}
 
 	pickedNodeName := hashRing.Get(strconv.Itoa(int(shardID)))
 	log.Debug("ConsistentHashNodePicker pick result", zap.Uint64("shardID", uint64(shardID)), zap.String("node", pickedNodeName), zap.Int("nodeNumber", aliveNodeNumber))
 
-	for _, registerNode := range registerNodes {
-		if registerNode.Node.Name == pickedNodeName {
-			return registerNode, nil
+	for i := 0; i < len(registerNodes); i++ {
+		if registerNodes[i].Node.Name == pickedNodeName {
+			return registerNodes[i], nil
 		}
 	}
 
--- a/server/coordinator/node_picker.go
+++ b/server/coordinator/node_picker.go
@@ -4,8 +4,6 @@ package coordinator
 
 import (
 	"context"
-	"crypto/rand"
-	"math/big"
 	"strconv"
 	"time"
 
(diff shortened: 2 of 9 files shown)


server/coordinator/node_picker.go
```diff
@@ -22,8 +25,9 @@ func NewRandomNodePicker() NodePicker {
 	return &RandomNodePicker{}
 }
 
-func (p *RandomNodePicker) PickNode(_ context.Context, registeredNodes []metadata.RegisteredNode) (metadata.RegisteredNode, error) {
-	now := time.Now().Unix()
+func (p *RandomNodePicker) PickNode(_ context.Context, _ storage.ShardID, registeredNodes []metadata.RegisteredNode) (metadata.RegisteredNode, error) {
```
ShiKaiWi, line 30: The `RandomNodePicker`'s pick logic is totally different from the `ConsistentNodePicker`, which may lead to a mass shard migration after the first shard assignment.

server/coordinator/node_picker.go
```diff
@@ -52,3 +58,33 @@ func (p *RandomNodePicker) PickNode(_ context.Context, registeredNodes []metadat
...
+	for _, registerNode := range registerNodes {
+		if !registerNode.IsExpired(now) {
+			nodeMapping[registerNode.Node.Name] = registerNode
+			hashRing.Add(registerNode.Node.Name)
+		}
+	}
+
+	log.Debug("pick node result", zap.Int("nodeNumber", len(nodeMapping)))
+	if len(nodeMapping) == 0 {
+		return metadata.RegisteredNode{}, errors.WithMessage(ErrNodeNumberNotEnough, "online node length must bigger than 0")
```
ShiKaiWi, line 84:
```suggestion
		return metadata.RegisteredNode{}, errors.WithMessage(ErrNodeNumberNotEnough, "at least one online nodes is required")
```

server/coordinator/node_picker.go
```diff
@@ -52,3 +58,33 @@ func (p *RandomNodePicker) PickNode(_ context.Context, registeredNodes []metadat
...
+	}
+
+	pickNodeName := hashRing.Get(strconv.Itoa(int(shardID)))
```
ShiKaiWi, line 87:
```suggestion
	pickedNodeName := hashRing.Get(strconv.Itoa(int(shardID)))
```

server/cluster/metadata/types.go
```diff
@@ -136,7 +136,7 @@ func NewRegisteredNode(meta storage.Node, shardInfos []ShardInfo) RegisteredNode
 }
 
 func (n RegisteredNode) IsExpired(now int64) bool {
-	return now >= int64(expiredThreshold)+int64(n.Node.LastTouchTime)
+	return now >= int64(expiredThreshold/time.Millisecond)+int64(n.Node.LastTouchTime)
```
ShiKaiWi, line 139: Shall we use the golang `time.Time` to do comparison?

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -60,7 +60,7 @@ func NewManager(procedureManager procedure.Manager, factory *coordinator.Factory
 		registerSchedulers: []Scheduler{},
 		isRunning:          false,
 		factory:            factory,
-		nodePicker:         coordinator.NewRandomNodePicker(),
+		nodePicker:         coordinator.NewConsistentHashNodePicker(5),
```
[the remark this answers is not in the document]
ZuLiangWang: Refer to the configuration of `groupcache` and set its default value to 50.

server/coordinator/node_picker.go
ZuLiangWang: It seems that `RandomNodePicker` is unnecessary in the case of `ConsistentHashNodePicker`, just remove it.

server/coordinator/node_picker.go
```diff
@@ -4,51 +4,47 @@ package coordinator
...
-	curOnlineIdx := -1
-	for idx := 0; idx < len(registeredNodes); idx++ {
-		if !registeredNodes[idx].IsExpired(now) {
-			curOnlineIdx++
-		}
-		if curOnlineIdx == selectIdx {
-			return registeredNodes[idx], nil
-		}
-	}
+	pickedNodeName := hashRing.Get(strconv.Itoa(int(shardID)))
```
ShiKaiWi, line 47: How about add debug log for the shardID and the pickedNodeName?

server/coordinator/node_picker.go
```diff
@@ -4,51 +4,47 @@ package coordinator
...
-	now := time.Now().Unix()
-	onlineNodeLength := 0
-	for _, registeredNode := range registeredNodes {
-		if !registeredNode.IsExpired(now) {
-			onlineNodeLength++
+func (p *ConsistentHashNodePicker) PickNode(_ context.Context, shardID storage.ShardID, registerNodes []metadata.RegisteredNode) (metadata.RegisteredNode, error) {
+	now := time.Now()
+
+	hashRing := hash.New(p.hashReplicas, nil)
+	nodeMapping := make(map[string]metadata.RegisteredNode, len(registerNodes))
```
ShiKaiWi, line 34: I guess there is no need to construct this map? With the hash result, we can find the node just by looping the slice without any memory allocation.

server/hash/consistent_hash_ring.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2023 CeresDB Project Authors. Licensed under Apache-2.0.
+// This file refer from [groupcache](https://github.com/golang/groupcache/blob/4a4ac3fbac33b83bb138f808c8945a2812023fc4/consistenthash/consistenthash.go)
```
ShiKaiWi, line 2:
```suggestion
// The consistent hash ring refers to [groupcache](https://github.com/golang/groupcache/blob/4a4ac3fbac33b83bb138f808c8945a2812023fc4/consistenthash/consistenthash.go)
```

ShiKaiWi (approved):
LGTM

PR #163, opened 2023-04-26
feat: add deploy mode

ZuLiangWang:

Closes #

# Rationale for this change
We support automatic failover and load balancing in the latest version, but they are not applicable in some scenarios, so add a switch to control their opening.

* Add `DeployMode` config, When the deployMode is turned on, the failover scheduling will be turned off, which is used for CeresDB cluster publishing and using local storage.

None.

# How does this change test
Pass all unit tests and verify that the switch takes effect in the local environment.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -47,7 +47,8 @@ const (
 	defaultClusterNodeCount         = 2
 	defaultClusterReplicationFactor = 1
 	defaultClusterShardTotal        = 8
-	defaultClusterDeployMode        = false
+	// TODO: DefaultClusterDeployMode should be set false, this is to be compatible with the implementation of CeresDB's local storage, which will be required later...
+	defaultClusterDeployMode = true
 
 	defaultHTTPPort = 8080
 
@@ -108,7 +109,7 @@ type Config struct {
 	DefaultClusterReplicationFactor int    `toml:"default-cluster-replication-factor" env:"DEFAULT_CLUSTER_REPLICATION_FACTOR"`
 	DefaultClusterShardTotal        int    `toml:"default-cluster-shard-total" env:"DEFAULT_CLUSTER_SHARD_TOTAL"`
 	// When the deployMode is turned on, the failover scheduling will be turned off, which is used for CeresDB cluster publishing and using local storage.
-	DefaultClusterDeployMode bool `toml:"default-cluster-deploy_mode" env:"DEFAULT_CLUSTER_DEPLOY_MODE"`
+	DefaultClusterDeployMode bool `toml:"default-cluster-deploy-mode" env:"DEFAULT_CLUSTER_DEPLOY_MODE"`
 
 	ClientUrls          string `toml:"client-urls" env:"CLIENT_URLS"`
 	PeerUrls            string `toml:"peer-urls" env:"PEER_URLS"`
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -43,11 +43,11 @@ const (
 	defaultMinScanLimit    int  = 20
 	defaultIDAllocatorStep uint = 20
 
-	defaultClusterName                     = "defaultCluster"
-	defaultClusterNodeCount                = 2
-	defaultClusterReplicationFactor        = 1
-	defaultClusterShardTotal               = 8
-	defaultPartitionTableProportionOfNodes = 0
+	defaultClusterName              = "defaultCluster"
+	defaultClusterNodeCount         = 2
+	defaultClusterReplicationFactor = 1
+	defaultClusterShardTotal        = 8
+	defaultClusterDeployMode        = false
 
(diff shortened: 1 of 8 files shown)


server/storage/types.go
```diff
@@ -21,6 +21,7 @@ type (
 const (
 	ClusterStateEmpty ClusterState = iota + 1
 	ClusterStateStable
+	ClusterStatePrepare
```
ShiKaiWi, line 24: Is this state not persisted?

server/service/http/api.go
```diff
@@ -53,6 +53,7 @@ func (a *API) NewAPIRouter() *Router {
 	router.Post("/route", a.route)
 	router.Post("/dropTable", a.dropTable)
 	router.Post("/getNodeShards", a.getNodeShards)
+	router.Post("/deployMode", a.deployMode)
```
ShiKaiWi, line 56: `PUT` method seems more suitable here?

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -182,3 +198,11 @@ func (m *ManagerImpl) Scheduler(ctx context.Context, clusterSnapshot metadata.Sn
 	}
 	return results
 }
+
+func (m *ManagerImpl) UpdateDeployMode(_ context.Context, deployMode bool) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	m.deployMode = deployMode
+	log.Info("scheduler manager update deploy mode", zap.Bool("deployMode", deployMode))
```
ShiKaiWi, line 207: Maybe we should log outside the lock here:
```suggestion
	m.lock.Lock()
	m.deployMode = deployMode
	m.lock.Unlock()


	log.Info("scheduler manager update deploy mode", zap.Bool("deployMode", deployMode))
```

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -110,6 +113,19 @@ func (m *ManagerImpl) Start(ctx context.Context) error {
 				// Get latest cluster snapshot.
 				clusterSnapshot := m.clusterMetadata.GetClusterSnapshot()
 				log.Debug("scheduler manager invoke", zap.String("clusterSnapshot", fmt.Sprintf("%v", clusterSnapshot)))
+
+				// TODO: Perhaps these codes related to deployMode need to be refactored.
+				// If deployMode is turned on, the scheduler will only be scheduled in the non-stable state.
+				if m.deployMode && clusterSnapshot.Topology.ClusterView.State == storage.ClusterStateStable {
+					continue
+				}
+				if clusterSnapshot.Topology.ClusterView.State == storage.ClusterStatePrepare && len(clusterSnapshot.Topology.ShardViewsMapping) == len(clusterSnapshot.Topology.ClusterView.ShardNodes) {
```
ShiKaiWi, line 122: Shall we add a method on Topology called `IsPrepareFinished` for this line code?

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -110,6 +113,19 @@ func (m *ManagerImpl) Start(ctx context.Context) error {
...
+					if err := m.clusterMetadata.UpdateClusterView(ctx, storage.ClusterStateStable, clusterSnapshot.Topology.ClusterView.ShardNodes); err != nil {
```
ShiKaiWi, line 123: It's weird to do cluster state update here? Is there any better place to do this?

server/config/config.go
```diff
@@ -43,11 +43,11 @@ const (
...
-	defaultClusterName                     = "defaultCluster"
-	defaultClusterNodeCount                = 2
-	defaultClusterReplicationFactor        = 1
-	defaultClusterShardTotal               = 8
-	defaultPartitionTableProportionOfNodes = 0
+	defaultClusterName              = "defaultCluster"
+	defaultClusterNodeCount         = 2
+	defaultClusterReplicationFactor = 1
+	defaultClusterShardTotal        = 8
+	defaultClusterDeployMode        = false
```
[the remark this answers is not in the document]
ZuLiangWang: Yes, it is safer to set it to true, I leave a todo here and raise an issue later.

server/storage/types.go
ZuLiangWang: I used a state that pb had before, which corresponds to `ClusterView_PREPARE_REBALANCE`.

server/coordinator/scheduler/scheduler_manager.go
ZuLiangWang: I also think the implementation here is a bit dirty, but for now, it can only be placed here. This corresponds to the todo I left above, and it will be modified when the code related to deployMode is refactored later.

ShiKaiWi (approved):
Good Job! 👍👍

PR #167, opened 2023-05-05
fix: fix drop partition table

ZuLiangWang:

Closes #

# Rationale for this change
Fixed the problem that drop partition table lacked parameters.

* Add `ClusterSnapshot` param to `DropPartitionTableProcedure`.
* Fixed the wrong judgment condition in `sampleCompare` and `sortCompare`.

None.

# How does this change test
Pass all unit tests.


chunshao90 (approved):
LGTM

PR #169, opened 2023-05-06
fix: fix compare result

ZuLiangWang:

Closes #

# Rationale for this change
Fix wrong return result in compare.

* Fix wrong return result in compare.

None.

# How does this change test
Pass all unit tests.


ShiKaiWi (approved):
LGTM

PR #170, opened 2023-05-06
fix: fix register node

ZuLiangWang:

Closes #

# Rationale for this change
Fix the problem of node registration.

* Fixed the problem where nodes were updated in the wrong order in `RegisterNode`.

None.

# How does this change test
Pass all unit tests.


ShiKaiWi (approved):
LGTM

PR #171, opened 2023-05-09
fix: mod zero

chunshao90:

Closes #

# Rationale for this change
 Fix the `% 0` bug that may occur during the cluster initialization phase.
 

*  Fix the `% 0` bug that may occur during the cluster initialization phase.
* Refactor `PickShards` function.
* Pick shard to random.

No.

# How does this change test
CI.

### DIFF
--- a/server/coordinator/shard_picker.go
+++ b/server/coordinator/shard_picker.go
@@ -14,11 +14,9 @@ import (
 
 // ShardPicker is used to pick up the shards suitable for scheduling in the cluster.
 // If expectShardNum bigger than cluster node number, the result depends on enableDuplicateNode:
-// If enableDuplicateNode is false, pick shards will be failed and return error.
-// If enableDuplicateNode is true, pick shard will return shards on the same node.
 // TODO: Consider refactor this interface, abstracts the parameters of PickShards as PickStrategy.
 type ShardPicker interface {
-	PickShards(ctx context.Context, snapshot metadata.Snapshot, expectShardNum int, enableDuplicateNode bool) ([]storage.ShardNode, error)
+	PickShards(ctx context.Context, snapshot metadata.Snapshot, expectShardNum int) ([]storage.ShardNode, error)
 }
 
 // RandomBalancedShardPicker randomly pick up shards that are not on the same node in the current cluster.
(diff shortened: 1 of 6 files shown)


server/coordinator/shard_picker.go
```diff
@@ -29,67 +27,68 @@ func NewRandomBalancedShardPicker() ShardPicker {
...
+				nodeNames = append(nodeNames, nodeName)
+			}
 		}
 
-		// Reset node shards when shard is all picked.
-		if len(result)%totalShardLength == 0 {
+		// Initialize tempNodeShardMapping in following cases:
+		// 1. First time to pick shards.
+		// 2. All shards are already picked.
+		if len(result) == 0 || len(result) == totalShardLength {
```
ZuLiangWang, line 59:
```suggestion
		if len(result)%totalShardLength == 0
```
We cannot use this judgment condition to judge whether the current shard has been fetched, for example, there are 4 shards now, and the goal is to pick 12 shards. We need to reset when 8 shards are picked, but the writing method here is unsatisfactory.

server/coordinator/shard_picker.go
```diff
@@ -29,67 +27,68 @@ func NewRandomBalancedShardPicker() ShardPicker {
 }
 
 // PickShards will pick a specified number of shards as expectShardNum.
-func (p *RandomBalancedShardPicker) PickShards(_ context.Context, snapshot metadata.Snapshot, expectShardNum int, enableDuplicateNode bool) ([]storage.ShardNode, error) {
+func (p *RandomBalancedShardPicker) PickShards(_ context.Context, snapshot metadata.Snapshot, expectShardNum int) ([]storage.ShardNode, error) {
 	shardNodes := snapshot.Topology.ClusterView.ShardNodes
 
 	nodeShardsMapping := make(map[string][]storage.ShardNode, 0)
```
ZuLiangWang, line 33: Add a conditional judgment here to avoid division by 0
```suggestion
	if len(shardNodes) == 0 {
	return storage.ShardNode{}, ErrShardNotEnough
	}
```

server/coordinator/shard_picker_test.go
```diff
@@ -19,25 +19,19 @@ func TestRandomShardPicker(t *testing.T) {
...
-
-	// Verify the number of shards and ensure that they are not on the same node.
-	re.Equal(len(shardNodes), 2)
-	re.NotEqual(shardNodes[0].NodeName, shardNodes[1].NodeName)
-
-	// ExpectShardNum is bigger than node number and enableDuplicateNode is false, it should be throw error.
-	_, err = shardPicker.PickShards(ctx, snapshot, 3, false)
-	re.Error(err)
 
 	// ExpectShardNum is bigger than node number and enableDuplicateNode is true, it should return correct shards.
```
ZuLiangWang, line 23: `enableDuplicatedNode` is removed now, correct old comment.

server/coordinator/shard_picker_test.go
```diff
@@ -19,25 +19,19 @@ func TestRandomShardPicker(t *testing.T) {
...
 	re.NoError(err)
 	re.Equal(len(shardNodes), 4)
 	// ExpectShardNum is bigger than shard number.
-	_, err = shardPicker.PickShards(ctx, snapshot, 5, true)
+	shardNodes, err = shardPicker.PickShards(ctx, snapshot, 5)
+	re.NoError(err)
+	re.Equal(len(shardNodes), 5)
+	shardNodes, err = shardPicker.PickShards(ctx, snapshot, 9)
 	re.NoError(err)
+	re.Equal(len(shardNodes), 9)
```
ZuLiangWang, line 36: Maybe we need add a check for shardNodes, we need to ensure that the shardNodes is average on nodes and shards.

chunshao90: Add `TODO`.

ZuLiangWang (approved):
LGTM

PR #172, opened 2023-05-09
feat: support multiple cluster

ZuLiangWang:

Closes #

# Rationale for this change
In order to avoid having to build a new CeresMeta cluster every time a new CeresDB cluster is created, we need to support the feature of one CeresMeta cluster managing multiple CeresDB clusters.

* Support create new CeresDB cluster & update cluster config & list clusters by http interface.
* Persist the enableSchedule and topologyType fields.
* To support cluster-level log output, create a child logger for each cluster.

None.

# How does this change test
Pass all unit tests and local manual tests.


server/storage/types.go
```diff
@@ -16,12 +16,16 @@ type (
...
 	NodeState    int
+	TopologyType string
 )
 
 const (
 	ClusterStateEmpty ClusterState = iota + 1
 	ClusterStateStable
 	ClusterStatePrepare
+
+	TopologyTypeStatic  = "static"
```
chunshao90, line 27: Use enum?

ZuLiangWang: `TopologyTypeStatic` need to be string, so `iota` is not used directly here.

server/service/http/api.go
```diff
@@ -367,30 +369,100 @@ func (a *API) split(writer http.ResponseWriter, req *http.Request) {
...
+	c, err := a.clusterManager.CreateCluster(req.Context(), createClusterRequest.ClusterName, metadata.CreateClusterOpts{
+		NodeCount:         createClusterRequest.ClusterNodeCount,
+		ReplicationFactor: 1,
+		ShardTotal:        createClusterRequest.ClusterShardTotal,
+		EnableSchedule:    createClusterRequest.EnableSchedule,
+		TopologyType:      topologyType,
+	})
+	if err != nil {
+		log.Error("create cluster", zap.Error(err))
+		a.respondError(writer, metadata.ErrCreateCluster, "create cluster failed")
```
chunshao90, line 426: log `err`.

server/service/http/api.go
```diff
@@ -367,30 +369,100 @@ func (a *API) split(writer http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		log.Error("decode request body failed", zap.Error(err))
+		a.respondError(writer, ErrParseRequest, "")
+		return
+	}
+
+	c, err := a.clusterManager.GetCluster(req.Context(), updateClusterRequest.ClusterName)
+	if err != nil {
+		log.Error("get cluster when update cluster", zap.Error(err))
+		a.respondError(writer, metadata.ErrClusterNotFound, fmt.Sprintf("cluster:%s not found", updateClusterRequest.ClusterName))
```
chunshao90, line 453:
```suggestion
		a.respondError(writer, metadata.ErrClusterNotFound, fmt.Sprintf("cluster not found, clusterName:%s", updateClusterRequest.ClusterName))
```

server/coordinator/scheduler/scheduler_manager.go
```diff
@@ -113,45 +114,44 @@ func (m *ManagerImpl) Start(ctx context.Context) error {
 		return errors.WithMessage(err, "start shard watch failed")
 	}
 
+	m.isRunning.Store(true)
```
chunshao90, line 117: Set in goroutine.

chunshao90 (approved):
LGTM

PR #177, opened 2023-05-16
chore: add integration test

ZuLiangWang:

Closes #

# Rationale for this change
CeresMeta is currently integrated in the cluster integration test of CeresDB. In order to avoid problems in CeresDB caused by changes in CeresMeta, the integration test of the cluster also needs to be introduced in CeresMeta.

* Add integration test in makefile.
* Update the workflow configuration and introduce integration tests into CI.

None.

# How does this change test
New workflow in this pull request could run normally.


run_integration_test.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -exo
+
+integration_test_path="/tmp/integration_tests"
```
chunshao90, line 5:
```suggestion
integration_test_path="/tmp/ceresmeta_integration_tests"
```

chunshao90 (approved):
LGTM

PR #179, opened 2023-05-19
fix: fix create table idempotence

ZuLiangWang:

Closes #

# Rationale for this change
In version 1.2.0, we have made a lot of modifications, resulting in the current table creation not having idempotent capabilities. If the table creation fails in CeresDB, it will lead to failure to build tables normally in the future. This problem is fixed in this pull request.

* Make table creation idempotent.
* Make `tableShardMapping` only mapping to single shard.

Users can retry to create tables that failed to create before.

# How does this change test

### DIFF
--- a/server/cluster/metadata/topology_manager.go
+++ b/server/cluster/metadata/topology_manager.go
@@ -23,6 +23,8 @@ type TopologyManager interface {
 	GetClusterState() storage.ClusterState
 	// GetTableIDs get shardNode and tablesIDs with shardID and nodeName.
 	GetTableIDs(shardIDs []storage.ShardID) map[storage.ShardID]ShardTableIDs
+	// GetShardIDs get the assigned shardID by tableID.
+	GetShardIDs(tableIDs []storage.TableID) map[storage.TableID]storage.ShardID
 	// AddTable add table to cluster topology.
 	AddTable(ctx context.Context, shardID storage.ShardID, tables []storage.Table) (ShardVersionUpdate, error)
 	// RemoveTable remove table on target shards from cluster topology.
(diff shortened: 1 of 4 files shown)


server/cluster/metadata/topology_manager.go
```diff
@@ -176,6 +178,21 @@ func (m *TopologyManagerImpl) GetTableIDs(shardIDs []storage.ShardID) map[storag
...
+func (m *TopologyManagerImpl) GetShardIDs(tableIDs []storage.TableID) map[storage.TableID]storage.ShardID {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+
+	tableShardMapping := make(map[storage.TableID]storage.ShardID, len(tableIDs))
+	for _, tableID := range tableIDs {
+		shardID, exists := m.tableShardMapping[tableID]
+		if exists {
+			tableShardMapping[tableID] = shardID
+		}
```
chunshao90, line 190: Return error if tableID not exists?

server/service/grpc/service.go
```diff
@@ -224,7 +224,7 @@ func (s *Service) DropTable(ctx context.Context, req *metaservicepb.DropTableReq
 		return ceresmetaClient.DropTable(ctx, req)
 	}
 
-	log.Info("[DropTable]", zap.String("schemaName", req.SchemaName), zap.String("clusterName", req.GetHeader().ClusterName), zap.String("tableName", req.Name))
+	log.Debug("[DropTable]", zap.String("schemaName", req.SchemaName), zap.String("clusterName", req.GetHeader().ClusterName), zap.String("tableName", req.Name))
```
chunshao90, line 227: `Create table` and `Drop table` log level should be info.
The log is very important.

server/cluster/metadata/topology_manager.go
ZuLiangWang: `TopologyManager.GetShardIDs` receive a `TableID` slice as parameter, if return error when some `tableID` not found in any shard, it means this function execute failed and we could not use it's execution result, this will make it impossible for us to judge which `TableID` exist on shards, I think the current implementation is a bit better.

PR #183, opened 2023-05-23
fix: add leader cache

### LINKED ISSUE #181: When there are a lot of route table requests, the CPU usage of the node is extremely high
**Description**
CeresMeta provides route table GRPC interface to CeresDB, we found that when there are a lot of route table requests, the CPU usage of the node is extremely high.

**Proposal**
Found the reason that the CPU usage of the node is extremely high, there may be hot code that needs to be optimized for performance.

**Additional context**

ZuLiangWang:

Closes #181 

# Rationale for this change
As mentioned in the issue, fix the performance problem of `routeTable`.

* Add leader memory cache in member.
* Correct the `GetLeader` method, get the leader endpoint through the cache when forwarding GRPC request.

None.

# How does this change test

### DIFF
--- a/server/member/member.go
+++ b/server/member/member.go
@@ -80,13 +80,19 @@ func (m *Member) getLeader(ctx context.Context) (*getLeaderResp, error) {
 	return &getLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetId() == m.ID}, nil
 }
 
-// GetLeader gets the leader of the cluster with memory cache.
+// GetLeaderAddr gets the leader address of the cluster with memory cache.
 // return error if no leader found.
-func (m *Member) GetLeader(_ context.Context) (*metastoragepb.Member, bool, error) {
+func (m *Member) GetLeaderAddr(_ context.Context) (GetLeaderAddrResp, error) {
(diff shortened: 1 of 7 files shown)


server/member/member.go
```diff
@@ -76,7 +77,16 @@ func (m *Member) GetLeader(ctx context.Context) (*GetLeaderResp, error) {
 		return nil, ErrInvalidLeaderValue.WithCause(err)
 	}
 
-	return &GetLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetId() == m.ID}, nil
+	return &getLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetId() == m.ID}, nil
+}
+
+// GetLeader gets the leader of the cluster with memory cache.
+// return error if no leader found.
+func (m *Member) GetLeader(_ context.Context) (*metastoragepb.Member, bool, error) {
```
ShiKaiWi, line 85: How about renaming this method to `GetLeaderAddr`, and returns a struct called LeaderAddr with only two fields:
- address of the leader
- boolean `Local`

ShiKaiWi (approved):
LGTM

PR #190, opened 2023-05-30
fix: remove test code

baojinri:

Closes #

# Rationale for this change
 Test case has a partial probability of failure.
 

remove partial probability of failure test code.

none

# How does this change test
CI


ZuLiangWang (approved):
LGTM

PR #191, opened 2023-05-30
refactor: compatible with topology persistence

ZuLiangWang:

Closes #

# Rationale for this change
We modified the persistent data `TopologyType` in the previous pull request. In order to avoid the user from manually operating the interface, we need to add compatible code

* Add compatible code for `TopologyType`

Avoid users manually operating the interface to be compatible with version changes

# How does this change test
An old version of the persistent data was used for testing in the local environment.

### DIFF
--- a/server/cluster/manager.go
+++ b/server/cluster/manager.go
@@ -56,9 +56,12 @@ type managerImpl struct {
 	alloc           id.Allocator
 	rootPath        string
 	idAllocatorStep uint
+
+	// TODO: topologyType is used to be compatible with cluster data changes and needs to be deleted later.
+	topologyType storage.TopologyType
 }
 
-func NewManagerImpl(storage storage.Storage, kv clientv3.KV, client *clientv3.Client, rootPath string, idAllocatorStep uint) (Manager, error) {
+func NewManagerImpl(storage storage.Storage, kv clientv3.KV, client *clientv3.Client, rootPath string, idAllocatorStep uint, topologyType storage.TopologyType) (Manager, error) {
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -161,7 +161,12 @@ func (srv *Server) startServer(_ context.Context) error {
 		MaxScanLimit: srv.cfg.MaxScanLimit, MinScanLimit: srv.cfg.MinScanLimit,
 	})
 
-	manager, err := cluster.NewManagerImpl(storage, srv.etcdCli, srv.etcdCli, srv.cfg.StorageRootPath, srv.cfg.IDAllocatorStep)
+	topologyType, err := metadata.ParseTopologyType(srv.cfg.TopologyType)
+	if err != nil {
+		return metadata.ErrParseTopologyType
+	}
```
chunshao90, line 167: return err.

server/cluster/manager.go
```diff
@@ -323,12 +327,34 @@ func (m *managerImpl) Start(ctx context.Context) error {
...
+					ShardTotal:     metadataStorage.ShardTotal,
+					EnableSchedule: metadataStorage.EnableSchedule,
+					TopologyType:   m.topologyType,
+					CreatedAt:      metadataStorage.CreatedAt,
+					ModifiedAt:     uint64(time.Now().UnixMilli()),
+				},
+			}); err != nil {
+				return errors.WithMessagef(err, "update cluster topology type failed, clusterName:%s", clusterMetadata.Name())
+			}
+			if err := clusterMetadata.LoadMetadata(ctx); err != nil {
```
chunshao90, line 347: Log update Cluster info.

chunshao90 (approved):
LGTM

Issue #199, opened 2023-06-06
Refactor the fsm of procedure


baojinri:
**Description**
Now, the ddl and operation fsm in procedure are not clearly, and the error messages of fsm are mixed together.

**Proposal**
- Sort out the state flow of ddl and operation
- Use CanceledError to distinct error message, refer: https://pkg.go.dev/github.com/looplab/fsm#CanceledError

**Additional context**

PR #201, opened 2023-06-07
refactor: fsm of create table

baojinri:

part of #199 

# Rationale for this change
 
 

# How does this change test


server/coordinator/procedure/ddl/createtable/create_table.go
```diff
@@ -190,17 +201,9 @@ func (p *Procedure) Start(ctx context.Context) error {
 		p:   p,
 	}
 
-	if err := p.fsm.Event(eventPrepare, req); err != nil {
-		err1 := p.fsm.Event(eventFailed, req)
+	if err := p.fsm.Event(eventCreateTable, req); err != nil {
```
ZuLiangWang, line 204: Maybe we can define a `FsmError` to judge whether it is an error in the internal execution of the state machine, so as to decide whether it can be retried.
```
if err := p.fsm.Event(eventCreateTable, req); err != nil {
switch err.(type){
 case: FsmError:
    if retrySize < MaxRetrySize{
       p.fsm.Event(eventCreateTable, req)
     }else{
      p.fsm.Event(eventRollback,req)
}
....
}
```

PR #202, opened 2023-06-12
fix: fix panic when shard is not assigned

ZuLiangWang:

Closes #

# Rationale for this change
Fix server panic when shard is not assigned.

* Add shardNodes check in `shard_picker.go`

None.

# How does this change test


chunshao90 (approved):
LGTM

PR #203, opened 2023-06-12
fix: add cluster state check

ZuLiangWang:

Closes #

# Rationale for this change
Before the cluster state is stable, it is not allowed to create and delete tables to avoid uneven distribution of tables.

* Add cluster state check when create table & drop table.

None.

# How does this change test


jiacai2050:
It's better to use interceptor to implements this, but this depend on https://github.com/etcd-io/etcd/issues/13468, so I plan to merge this first.

jiacai2050 (approved):
LGTM

PR #207, opened 2023-06-27
feat: add ETCD member API

### LINKED ISSUE #209: Support to dynamically modify peers and leader of CeresMeta cluster through the interface
**Description**
At present, the cluster configuration of `CeresMeta` only supports modification in the form of configuration items, and it cannot be modified after the cluster initialization is completed. When the `CeresMeta` cluster nodes are added or deleted, the operation and maintenance is very complicated. We need to provide a set of interfaces to support the cluster Dynamic modification of peer and leader.

**Proposal**
Support to dynamically modify peers and leader of CeresMeta cluster through the interface.

**Additional context**

ZuLiangWang:

Closes #209 

# Rationale for this change
At present, the peers IP of CeresMeta cannot be adjusted after the cluster initialization is completed, and the capacity cannot be expanded or contracted. If there is a node replacement or expansion or contraction, it needs to be manually operated through `etcdutil`. We need to add a management interface to solve this problem.

* Add ETCD member API. 

None.

# How does this change test

### DIFF
--- a/server/service/http/etcd_api.go
+++ b/server/service/http/etcd_api.go
@@ -25,7 +25,7 @@ func NewEtcdAPI(etcdClient *clientv3.Client, forwardClient *ForwardClient) EtcdA
 }
 
 type AddMemberRequest struct {
-	MembersAddr []string `json:"membersAddr"`
+	MemberAddrs []string `json:"memberAddrs"`
 }
 
 func (a *EtcdAPI) addMember(writer http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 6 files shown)


server/service/http/etcd_api.go
```diff
@@ -0,0 +1,215 @@
...
+
+func NewEtcdAPI(etcdClient *clientv3.Client, forwardClient *ForwardClient) EtcdAPI {
+	return EtcdAPI{
+		etcdClient:    etcdClient,
+		forwardClient: forwardClient,
+	}
+}
+
+type AddMemberRequest struct {
+	MembersAddr []string `json:"membersAddr"`
```
jiacai2050, line 28: memberAddrs

server/service/http/etcd_api.go
```diff
@@ -0,0 +1,215 @@
...
+			_, err := a.etcdClient.MemberUpdate(req.Context(), member.ID, updateMemberRequest.NewMemberAddr)
+			if err != nil {
+				log.Error("remove learner failed", zap.Error(err))
+				respondError(writer, ErrRemoveMembers, fmt.Sprintf("err: %s", err.Error()))
+				return
+			}
+			respond(writer, "ok")
+			return
+		}
+	}
```
jiacai2050, line 93: return err when no members found?

jiacai2050 (approved):
LGTM

PR #212, opened 2023-07-05
fix: fix codecov

ZuLiangWang:

Closes #

# Rationale for this change
Fix the code cov problem, there is a problem with the current baseline, the expired code is used, the baseline is not updated.

* Fix code cov config.

None.

# How does this change test
No need to test.


jiacai2050 (approved):
LGTM

Issue #208, opened 2023-06-28
Optimize the hash algorithm to make the distribution of shards and tables more even


ZuLiangWang:
**Description**
We currently use a simple version of consistent hashing as the core algorithm of the scheduling strategy. It cannot guarantee the uniform distribution of tables and shards. We need to optimize this basic version to make the scheduling strategy more perfect.

**Proposal**
Optimize the hash algorithm to make the distribution of shards and tables more even.

**Additional context**

PR #215, opened 2023-07-12
refactor: optimize hash node select algo

### LINKED ISSUE #200: Balance  the shard distribution on servers
**Description**

Now the distribution of shard between servers may be not even, which cause high load on some server while low load on others. For example in my cluster, there are 128 shards and 30 servers, the shard count in every server like this:
```
4 
2 
8 
3 
7 
7 
3 
7 
2 
6 
3 
6 
4 
4 
7 
7 
3 
2 
3 
3 
5 
4 
2 
4 
3 
3 
4 
2 
6 
4 
```
We can observe the following:

* The maximum count of shards on a single server is 8.
* The minimum count of shards on a single server is 2.

**Proposal**

Make the distribution of shards more even.

**Additional context**

ZuLiangWang:

Closes #200  #208 

# Rationale for this change
As mentioned in the issue, we need to optimize the current algorithm for assigning shards to nodes so that the distribution of shards on nodes becomes more even.

* Replace `ConsistentHashRing` with `UniformityHash`.
* Refactor `NodePicker` based on `uniformityHash`, which make the distribution of shards on nodes becomes more even.

* The distribution of shards on nodes becomes more even.

# How does this change test
Pass all unit tests and integration test, and observe the distribution of Shard in the unit test to ensure that its distribution is reasonable.


server/coordinator/node_picker.go
```diff
@@ -4,53 +4,78 @@ package coordinator
...
-type ConsistentHashNodePicker struct {
-	logger       *zap.Logger
-	hashReplicas int
+type UniformityConsistentHashNodePicker struct {
+	logger *zap.Logger
 }
 
-func NewConsistentHashNodePicker(logger *zap.Logger, hashReplicas int) NodePicker {
-	return &ConsistentHashNodePicker{hashReplicas: hashReplicas, logger: logger}
+func NewUniformityConsistentHashNodePicker(logger *zap.Logger) NodePicker {
```
chunshao90, line 25: The schedule can be placed on the cluster, so that different clusters can support different scheduling algorithms.

server/coordinator/node_picker.go
```diff
@@ -4,53 +4,91 @@ package coordinator
...
+	}
+	c := hash.NewUniformityHash(mems, consistentConfig)
+	for partID := 0; partID < consistentConfig.PartitionCount; partID++ {
+		mem := c.GetPartitionOwner(partID)
+		nodeName := mem.String()
 
-	for i := 0; i < len(registerNodes); i++ {
-		if registerNodes[i].Node.Name == pickedNodeName {
-			return registerNodes[i], nil
+		if contains(storage.ShardID(partID), shardIDs) {
```
jiacai2050, line 70: When partID will not be in shardIDs?

server/coordinator/node_picker.go
```diff
@@ -4,53 +4,91 @@ package coordinator
...
+func NewUniformityConsistentHashNodePicker(logger *zap.Logger) NodePicker {
+	return &UniformityConsistentHashNodePicker{logger: logger}
 }
 
-func NewConsistentHashNodePicker(logger *zap.Logger, hashReplicas int) NodePicker {
-	return &ConsistentHashNodePicker{hashReplicas: hashReplicas, logger: logger}
+type nodeMember string
+
+func (m nodeMember) String() string {
+	return string(m)
```
jiacai2050, line 31:
```suggestion
	return m
```

server/coordinator/node_picker.go
```diff
@@ -4,53 +4,91 @@ package coordinator
...
-type ConsistentHashNodePicker struct {
-	logger       *zap.Logger
-	hashReplicas int
+func NewUniformityConsistentHashNodePicker(logger *zap.Logger) NodePicker {
+	return &UniformityConsistentHashNodePicker{logger: logger}
 }
 
-func NewConsistentHashNodePicker(logger *zap.Logger, hashReplicas int) NodePicker {
-	return &ConsistentHashNodePicker{hashReplicas: hashReplicas, logger: logger}
+type nodeMember string
```
jiacai2050, line 28: why this new type is necessary?

server/coordinator/node_picker.go
ZuLiangWang: This modification will cause compilation errors, golang compiler seems to be unable to recognize type alias normally.

server/coordinator/node_picker.go
ZuLiangWang: `UniformityHash` receives a parameter of type member, the `NodeMember` implements two interfaces of member and string at the same time.

server/coordinator/node_picker.go
ZuLiangWang: It is impossible in theory, but here to be on the safe side, a check is done.

jiacai2050 (approved):
LGTM

PR #218, opened 2023-07-18
feat: add reopen shard scheduler

ZuLiangWang:

Closes #

# Rationale for this change
At present, our implementation of shard opening is relatively simple. When some tables in a shard fail to open, it is impossible to retry and recover without restarting the CeresDB node. We need to solve this problem so that shards that fail to open some tables Can be recovered by retrying.

* Introduce the latest proto and add the definition of shard status.
* Add the implementation of `ReopenShardScheduler` for retrying partially opened shards.

None.

# How does this change test
* Pass all unit tests and integration tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.19
 
 require (
 	github.com/AlekSi/gocov-xml v1.0.0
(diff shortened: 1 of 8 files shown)


jiacai2050 (approved):
LGTM

PR #220, opened 2023-07-21
refactor: refactor api module

ZuLiangWang:

Closes #

# Rationale for this change
Refactor the api module and encapsulate the repeated code as a wrapper.

* Add `wrap` function in `api.go`, it is used to encapsulate result processing logic and request forwarding logic.
* Refactor api module based on `wrap` function.

None.

# How does this change test

### DIFF
--- a/server/service/http/api.go
+++ b/server/service/http/api.go
@@ -39,6 +39,22 @@ type apiFuncResult struct {
 	errMsg string
 }
 
+func okResult(data interface{}) apiFuncResult {
+	return apiFuncResult{
+		data:   data,
+		err:    nil,
+		errMsg: "",
+	}
+}
+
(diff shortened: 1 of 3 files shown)


server/service/http/api.go
```diff
@@ -33,6 +33,14 @@ const (
 	apiPrefix string = "/api/v1"
 )
 
+type apiFuncResult struct {
+	data   interface{}
+	err    coderr.CodeError
+	errMsg string
+}
+
+type apiFunc func(r *http.Request) apiFuncResult
```
jiacai2050, line 42: Define two help method:
```go
func okResult(data interface{}) apiFuncResult
func errResult(err coderr.CodeError, msg:String) apiFuncResult
```

jiacai2050 (approved):
LGTM

PR #222, opened 2023-07-27
fix: remove gprc header error stack

### LINKED ISSUE #221: The CPU utilization rate is extremely high when there are many route requests
**Describe this problem**
At present, we found that when the CeresDB node is restarted, a large number of route requests will cause the CPU utilization of CeresMeta to soar, and a large number of request timeouts will occur. By observing the flame graph, we find that the formatting of the exception stack will consume a lot of CPU calculations. We need Optimize it to solve the problem of abnormally high CPU rate usage.

**Steps to reproduce**

**Expected behavior**

**Additional Information**
截屏2023-07-27 14 41 59

ZuLiangWang:

Closes #221 

# Rationale for this change
As mentioned in issue, reduce the pressure of route requests on cpu.

* Remove error stack in gprc response header.

None.

# How does this change test
No need.

### DIFF
--- a/server/service/grpc/service.go
+++ b/server/service/grpc/service.go
@@ -409,10 +409,10 @@ func responseHeader(err error, msg string) *commonpb.ResponseHeader {
 
 	code, ok := coderr.GetCauseCode(err)
 	if ok {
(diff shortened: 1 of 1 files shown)


jiacai2050 (approved):
LGTM

PR #230, opened 2023-09-12
fix: wrong shard id when drop non-exist table

ShiKaiWi:

Closes #

# Rationale for this change
 The shard is wrong when dropping a non-existent table.
 

Throw an error when the table to drop is not found.

None.

# How does this change test
Manually.

### DIFF
--- a/server/coordinator/factory.go
+++ b/server/coordinator/factory.go
@@ -187,6 +187,10 @@ func (f *Factory) makeCreatePartitionTableProcedure(ctx context.Context, request
 	})
 }
 
+// CreateDropTableProcedure creates a procedure to do drop table.
+//
+// And if no error is thrown, the returned boolean value is used to tell whether the procedure is created.
+// In some cases, e.g. the table doesn't exist, it should not be an error and false will be returned.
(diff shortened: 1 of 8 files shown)


ZuLiangWang (approved):
LGTM

server/coordinator/factory.go
```diff
@@ -187,10 +187,14 @@ func (f *Factory) makeCreatePartitionTableProcedure(ctx context.Context, request
...
-func (f *Factory) CreateDropTableProcedure(ctx context.Context, request DropTableRequest) (procedure.Procedure, error) {
+// CreateDropTableProcedure creates a procedure to do drop table.
+//
+// And if no error is thrown, the returned boolean value is used to tell whether the procedure is created.
+// In some cases, e.g. the table doesn't exist, it should not be an error and false will be returned.
+func (f *Factory) CreateDropTableProcedure(ctx context.Context, request DropTableRequest) (procedure.Procedure, bool, error) {
 	id, err := f.allocProcedureID(ctx)
 	if err != nil {
-		return nil, err
+		return nil, false, err
```
ZuLiangWang, line 197:
```suggestion
		return nil, true, err
```
Returning false means that this error will not be passed to the user. If alloc table id fails, we need to pass this error out.

ZuLiangWang (approved):
LGTM

PR #236, opened 2023-09-21
improve: generate a more uniform distribution

ShiKaiWi:
## Rationale
Currently, the shard distribution over the registered nodes is not uniform, that is to say, some nodes may be idle while some ones may be very busy.

## Detailed Changes
Refactor the consistent-uniform hash implementation to make it real uniform, and surely it may be a little bit less consistent compared with before.

## Test Plan
Add the unit tests for the uniform and consistency.


server/coordinator/node_picker.go
```diff
@@ -33,64 +35,61 @@ func (m nodeMember) String() string {
...
-				}
-			}
-		}
-	}
-
-	if len(result) != len(shardIDs) {
-		return map[storage.ShardID]metadata.RegisteredNode{}, errors.WithMessagef(ErrPickNode, "length of the result is inconsistent with the length of the shard, shardID:%d, shardTotalNum:%d, aliveNodeNum:%d", shardIDs, shardTotalNum, len(aliveNodes))
+	h, err := hash.BuildConsistentUniformHash(int(shardTotalNum), mems, conf)
+	if err != nil {
+		return nil, err
```
chunshao90, line 79: return nil, errors.WithMessage(ErrPickNode, "...")

chunshao90 (approved):
LGTM

PR #238, opened 2023-09-29
feat: support external etcd

chunshao90:
## Rationale
Support external etcd cluster.

## Detailed Changes
* Add option to enable embed etcd cluster.
* [Breaking change] rename `default-http-port` to `http-port`.

## Test Plan
Manual test and CI.

### DIFF
--- a/server/member/member.go
+++ b/server/member/member.go
@@ -77,7 +77,7 @@ func (m *Member) getLeader(ctx context.Context) (*getLeaderResp, error) {
 		return nil, ErrInvalidLeaderValue.WithCause(err)
 	}
 
-	return &getLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetId() == m.ID}, nil
+	return &getLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetEndpoint() == m.Endpoint}, nil
 }
 
 // GetLeaderAddr gets the leader address of the cluster with memory cache.
@@ -91,7 +91,7 @@ func (m *Member) GetLeaderAddr(_ context.Context) (GetLeaderAddrResp, error) {
 	}
 	return GetLeaderAddrResp{
 		LeaderEndpoint: m.leader.Endpoint,
-		IsLocal:        m.leader.GetId() == m.ID,
+		IsLocal:        m.leader.Endpoint == m.Endpoint,
 	}, nil
 }
 
--- a/server/member/member.go
+++ b/server/member/member.go
@@ -229,7 +229,10 @@ func (m *Member) CampaignAndKeepLeader(ctx context.Context, leaseTTLSec int64, e
 
 			if enableEmbedEtcd {
 				// Check the etcd leader periodically. If the etcd leader is changed, ceresmeta leader should be re-elected.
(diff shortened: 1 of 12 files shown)


server/server.go
```diff
@@ -268,7 +330,10 @@ func (ctx *leaderWatchContext) ShouldStop() bool {
 }
 
 func (ctx *leaderWatchContext) EtcdLeaderID() uint64 {
-	return ctx.srv.etcdSrv.Server.Lead()
+	if ctx.srv.etcdSrv != nil {
+		return ctx.srv.etcdSrv.Server.Lead()
+	}
+	return 0
```
jiacai2050, line 336: what does 0 mean?

server/member/member.go
```diff
@@ -77,7 +77,7 @@ func (m *Member) getLeader(ctx context.Context) (*getLeaderResp, error) {
 		return nil, ErrInvalidLeaderValue.WithCause(err)
 	}
 
-	return &getLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetId() == m.ID}, nil
+	return &getLeaderResp{Leader: leader, Revision: leaderKv.ModRevision, IsLocal: leader.GetEndpoint() == m.Endpoint}, nil
```
jiacai2050, line 80: - Add a method `isLocal` to encapsulate this logic.
- Why change id check to endpoint?

server/member/watch_leader.go
```diff
@@ -26,25 +26,35 @@ type WatchContext interface {
 }
 
 type LeaderWatcher struct {
-	watchCtx    WatchContext
-	self        *Member
-	leaseTTLSec int64
+	watchCtx        WatchContext
+	self            *Member
+	leaseTTLSec     int64
+	watchEtcdLeader bool
```
jiacai2050, line 32: Add comments explaining when this will be false/true.

server/member/watch_leader.go
```diff
@@ -131,3 +141,65 @@ func (l *LeaderWatcher) Watch(ctx context.Context, callbacks LeadershipEventCall
...
+		}
+
+		if wait != waitReasonNoWait {
+			logger.Warn("sleep a while during watch", zap.String("wait-reason", wait))
+			time.Sleep(watchLeaderFailInterval)
+			wait = waitReasonNoWait
+		}
+
+		// Check whether leader exists.
+		leaderResp, err := l.self.getLeader(ctx)
```
jiacai2050, line 178: resp

server/server.go
```diff
@@ -89,9 +98,18 @@ func (srv *Server) Run(ctx context.Context) error {
 		return err
 	}
 
+	// If enableEmbedEtcd is false, the grpc server is started in a separate process.
+	if !srv.cfg.EnableEmbedEtcd {
```
jiacai2050, line 102: Remove this if block, add start grpc server at L90 with a else branch.

server/member/member.go
chunshao90: The original implementation uses etcd id to determine whether it is local node. Now when using an external etcd cluster, you need to judge by the registered node address.

server/server.go
chunshao90: return err.

ShiKaiWi (approved):

PR #240, opened 2023-09-30
feat: support drop partition table parallelly

chunshao90:
## Rationale
Support concurrent deletion of partition subtables at the shard level.

## Detailed Changes
* Dispatch drop partition sub table concurrently.

## Test Plan
Manual test and CI.
```
CREATE TABLE `demo1`(
    `name`string TAG,
    `id` int TAG,
    `value` double NOT NULL,
    `t` timestamp NOT NULL,
    TIMESTAMP KEY(t)
    ) PARTITION BY KEY(name) PARTITIONS 400 ENGINE = Analytic
```
It takes 4 seconds to drop this partition table `demo1` which contains 400 sub-tables.

### DIFF
--- a/server/coordinator/procedure/ddl/droppartitiontable/drop_partition_table.go
+++ b/server/coordinator/procedure/ddl/droppartitiontable/drop_partition_table.go
@@ -285,7 +285,7 @@ func dropDataTablesCallback(event *fsm.Event) {
 	g, _ := errgroup.WithContext(req.ctx)
 
 	// shardID -> tableNames
-	splitedTableNames := make(map[storage.ShardID][]string)
+	shardTables := make(map[storage.ShardID][]string)
 	for _, tableName := range params.SourceReq.PartitionTableInfo.GetSubTableNames() {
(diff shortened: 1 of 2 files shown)


server/coordinator/procedure/ddl/droppartitiontable/drop_partition_table.go
```diff
@@ -281,7 +282,11 @@ func dropDataTablesCallback(event *fsm.Event) {
 	}
 
 	shardVersions := req.p.relatedVersionInfo.ShardWithVersion
-	for _, tableName := range params.SourceReq.PartitionTableInfo.SubTableNames {
+	g, _ := errgroup.WithContext(req.ctx)
+
+	// shardID -> tableNames
+	splitedTableNames := make(map[storage.ShardID][]string)
```
ShiKaiWi, line 288:
```suggestion
	shardTables := make(map[storage.ShardID][]string)
```

ShiKaiWi (approved):

PR #245, opened 2023-10-03
feat: support shard affinity

ShiKaiWi:
## Rationale
In some cases, some specific shards may contain tables whose query or write requests are massive, and these tables are hoped to be migrated to some node alone.

## Detailed Changes
- Provide the shard affinity mechanism when generating the shard topology
- Support to manipulate the affinity configures by debug api
- Regroupping the packages of different scheduler

## Test Plan
New unit tests are added.


server/coordinator/scheduler/manager/scheduler_manager.go
```diff
@@ -260,13 +279,51 @@ func (m *ManagerImpl) UpdateDeployMode(ctx context.Context, enable bool) error {
...
 	if m.topologyType != storage.TopologyTypeDynamic {
-		return false, errors.WithMessage(ErrInvalidTopologyType, "deploy mode could only get when topology type is dynamic")
+		return false, ErrInvalidTopologyType.WithCausef("deploy mode could only get when topology type is dynamic")
 	}
 
 	return m.deployMode, nil
 }
+
+func (m *schedulerManagerImpl) AddShardAffinityRule(ctx context.Context, rule scheduler.ShardAffinityRule) error {
+	var firstErr error
```
ZuLiangWang, line 294:
```suggestion
	var lastErr error
```

server/coordinator/scheduler/manager/scheduler_manager.go
```diff
@@ -260,13 +279,51 @@ func (m *ManagerImpl) UpdateDeployMode(ctx context.Context, enable bool) error {
...
+	for _, scheduler := range m.registerSchedulers {
+		if err := scheduler.AddShardAffinityRule(ctx, rule); err != nil {
+			firstErr = err
+		}
+	}
+
+	return firstErr
```
ZuLiangWang, line 301: Add some comment to explain the meaning of error.

server/coordinator/scheduler/manager/scheduler_manager.go
```diff
@@ -260,13 +279,51 @@ func (m *ManagerImpl) UpdateDeployMode(ctx context.Context, enable bool) error {
...
 	}
 
 	return m.deployMode, nil
 }
+
+func (m *schedulerManagerImpl) AddShardAffinityRule(ctx context.Context, rule scheduler.ShardAffinityRule) error {
+	var firstErr error
+	for _, scheduler := range m.registerSchedulers {
+		if err := scheduler.AddShardAffinityRule(ctx, rule); err != nil {
+			firstErr = err
```
ZuLiangWang, line 297:
```suggestion
			firstErr = err
```
Add some log get error msg.

server/coordinator/scheduler/manager/scheduler_manager.go
```diff
@@ -260,13 +279,51 @@ func (m *ManagerImpl) UpdateDeployMode(ctx context.Context, enable bool) error {
...
+		}
+	}
+
+	return firstErr
+}
+
+func (m *schedulerManagerImpl) RemoveShardAffinityRule(ctx context.Context, shardID storage.ShardID) error {
```
ZuLiangWang, line 304: Same as `AddShardAffinityRule`.

server/coordinator/scheduler/manager/scheduler_manager.go
```diff
@@ -260,13 +279,51 @@ func (m *ManagerImpl) UpdateDeployMode(ctx context.Context, enable bool) error {
...
+	for _, scheduler := range m.registerSchedulers {
+		if err := scheduler.RemoveShardAffinityRule(ctx, shardID); err != nil {
+			firstErr = err
+		}
+	}
+
+	return firstErr
+}
+
+func (m *schedulerManagerImpl) ListShardAffinityRules(ctx context.Context) (map[string]scheduler.ShardAffinityRule, error) {
```
ZuLiangWang, line 315: Same as `AddShardAffinityRule`.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -5,7 +5,7 @@
 //
 // This code is licensed under the MIT License.
 //
-// Permission is hereby granted, free of charge, to any person obtaining a copy
+// Permission is hereby granted, free G charge, to any person obtaining a copy
```
ZuLiangWang, line 8:
```suggestion
// Permission is hereby granted, free of charge, to any person obtaining a copy
```

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -86,14 +90,14 @@ type virtualNode uint64
...
-	maxLoad       float64
+	minLoad       int
+	maxLoad       int
 	numPartitions uint32
 	// Member name => Member
 	members map[string]Member
-	// Member name => Member's load
-	memLoads map[string]float64
-	// partition id => index of the virtualNode in the sortedRing
+	// Member name => Partition ID
```
ZuLiangWang, line 98: This line of comments does not seem to correspond to the code. Why the following code is `map[int]struct`?

server/coordinator/scheduler/rebalanced/scheduler.go
```diff
@@ -28,45 +31,77 @@ type RebalancedShardScheduler struct {
 	// when deployMode is true, rebalancedShardScheduler will recover cluster according to the topology.
 	latestShardNodeMapping map[storage.ShardID]metadata.RegisteredNode
 	deployMode             bool
+	shardAffinityRule      map[storage.ShardID]scheduler.ShardAffinity
```
ZuLiangWang, line 34: Add some comment to explain shardAffinityRule.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -60,6 +63,8 @@ var (
 	ErrEmptyMembers = errors.New("at least one member is required")
 )
 
+const hashSeparator = "@$"
```
ZuLiangWang, line 66: What does this hashSeparator do? Maybe you need to add a comment to explain.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -195,20 +213,52 @@ func (c *ConsistentUniformHash) distributePartitions() {
...
 	for _, mem := range members {
 		for i := 0; i < c.config.ReplicationFactor; i++ {
 			// TODO: Shall use a more generic hasher which receives multiple slices or string?
-			key := []byte(fmt.Sprintf("%s%d", mem.String(), i))
+			key := []byte(fmt.Sprintf("%s%s%d", mem.String(), hashSeparator, i))
 			h := virtualNode(c.config.Hasher.Sum64(key))
+
+			oldMem, ok := c.nodeToMems[h]
+			if ok {
+				log.Warn("found has collision", zap.String("oldMem", oldMem.String()), zap.String("newMem", mem.String()))
```
ZuLiangWang, line 259:
```suggestion
				log.Warn("found has hash collision", zap.String("oldMem", oldMem.String()), zap.String("newMem", mem.String()))
```

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -195,20 +224,52 @@ func (c *ConsistentUniformHash) distributePartitions() {
...
+	// Ensure the order of members to avoid inconsistency caused by hash collisions.
+	sort.Slice(members, func(i, j int) bool {
+		return members[i].String() < members[j].String()
+	})
+
 	for _, mem := range members {
 		for i := 0; i < c.config.ReplicationFactor; i++ {
 			// TODO: Shall use a more generic hasher which receives multiple slices or string?
-			key := []byte(fmt.Sprintf("%s%d", mem.String(), i))
+			key := []byte(fmt.Sprintf("%s%s%d", mem.String(), hashSeparator, i))
```
ZuLiangWang, line 265: Why use `hashSeparator`? Can it reduce the probability of hash collision?

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -220,24 +281,96 @@ func (c *ConsistentUniformHash) initializeVirtualNodes(members []Member) {
...
 	}
-	return res
 }
 
-// GetPartitionOwner returns the owner of the given partition.
-func (c *ConsistentUniformHash) GetPartitionOwner(partID int) Member {
-	virtualNodeIdx, ok := c.partitionDist[partID]
-	if !ok {
-		return nil
+// offloadMember tries to offload the given member by moving the its partitions to another members.
```
ZuLiangWang, line 305:
```suggestion
// offloadMember tries to offload the given member by moving the partitions to other members.
```

config/example-standalone.toml
```diff
@@ -7,6 +7,7 @@ data-dir = "/tmp/ceresmeta0"
 node-name = "meta0"
 initial-cluster = "meta0=http://127.0.0.1:2380"
 default-cluster-node-count = 2
+topology-type = "dynamic"
```
ZuLiangWang, line 10: Reset this config.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -220,24 +281,96 @@ func (c *ConsistentUniformHash) initializeVirtualNodes(members []Member) {
...
+	for load := c.maxLoad; load < int(loadUpperBound); load++ {
+		if done := c.offloadPartitionWithAllowedLoad(sourcePartID, sourceMem, load, blackedMembers); done {
+			break
+		}
+	}
+}
+
+func (c *ConsistentUniformHash) offloadPartitionWithAllowedLoad(sourcePartID int, sourceMem Member, allowedMaxLoad int, blackedMembers map[string]struct{}) bool {
+	vNodeIdx := c.partitionDist[sourcePartID]
+	// Skip the first member which must not be the target to move.
```
ZuLiangWang, line 344: I can't understand, why do you need to skip the first member?

server/coordinator/scheduler/manager/scheduler_manager.go
ShiKaiWi: I guess there is no need to print the error on this level?

server/coordinator/scheduler/manager/scheduler_manager.go
ShiKaiWi: I can't figure out how to give comments to an implicit error, could you provide some suggestions about such comments?

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
ShiKaiWi: Make it hard to build the duplicate keys here.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
```diff
@@ -126,24 +140,37 @@ func BuildConsistentUniformHash(numPartitions int, members []Member, config Conf
...
+		memPartitions: memPartitions,
 		members:       make(map[string]Member, len(members)),
-		memLoads:      make(map[string]float64, len(members)),
 		partitionDist: make(map[int]int, numPartitions),
 		nodeToMems:    make(map[virtualNode]Member, numReplicatedNodes),
 	}
 
 	c.initializeVirtualNodes(members)
 	c.distributePartitions()
+	c.ensureAffinity()
```
ZuLiangWang, line 171: What will happen If the affinity rules cannot be satisfied?
For example:
There are now 10 shards and 2 nodes, but the NumAllowedOtherPartitions of 3 shards are set to 0.
Is it necessary to do a pre-verification of affinity rules?

server/coordinator/scheduler/manager/scheduler_manager.go
ZuLiangWang: The logic of error processing here is similar to errorgroup. I intercept the code in errorgroup.
```
// Go calls the given function in a new goroutine.
//
// The first call to return a non-nil error cancels the group; its error will be
// returned by Wait.
func (g *Group) Go(f func() error) {
	g.wg.Add(1)

	go func() {
		defer g.wg.Done()

		if err := f(); err != nil {
			g.errOnce.Do(func() {
				g.err = err
				if g.cancel != nil {
					g.cancel()
				}
			})
		}
	}()
}
```
Add a description of the error returned, like this:
```
// The first call to return a non-nil error cancels the group; its error will be
// returned by Wait.
```

server/coordinator/scheduler/manager/scheduler_manager.go
ZuLiangWang: Yes, waring or info is enough, what I mean is that just adding a line of log here.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
ShiKaiWi: Actually, some affinity rule may not be met if so, but it will stop.

server/coordinator/scheduler/nodepicker/hash/consistent_uniform.go
ShiKaiWi: Because the `sourcePartitionID` belongs to the firstMemer (`sourceMem`).

ZuLiangWang (approved):
LGTM

PR #247, opened 2023-10-07
feat: add deploy mode

ZuLiangWang:
## Rationale
In the current implementation, every time the CeresDB node changes, it will cause large-scale scheduling in the cluster. However, in the actual situation, we hope to control the affected range of the cluster when deploy and debug. Therefore, we need to add a deploy mode switch to control cluster scheduling.

## Detailed Changes
* Add `DeployMode` switch in `RebalancedShardScheduler`, shard topology will be locked when deploy mode is true.
* Add some API to control `DeployMode`.

## Test Plan
Pass all unit tests and test `DeployMode` take effect locally.

### DIFF
--- a/server/coordinator/scheduler/error.go
+++ b/server/coordinator/scheduler/error.go
@@ -1,0 +1,9 @@
+// Copyright 2023 CeresDB Project Authors. Licensed under Apache-2.0.
+
+package scheduler
+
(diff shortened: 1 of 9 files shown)


ShiKaiWi (approved):
LGTM

PR #248, opened 2023-10-08
feat: add least table shard picker

ZuLiangWang:
## Rationale
Currently, we use a completely random method to select shards for tables, which will lead to uneven distribution of tables in the cluster when the number of tables is small. In order to solve this problem, a new shardPicker is implemented, which selects the shard with the smallest number of current tables each time to create a table.

## Detailed Changes
* Add `NewLeastTableShardPicker`, which selects the shard with the smallest number of current tables each time to create a table.

## Test Plan
* Pass all unit tests and manually create tables for verification in the local environment.

### DIFF
--- a/server/coordinator/shard_picker.go
+++ b/server/coordinator/shard_picker.go
@@ -4,8 +4,6 @@ package coordinator
 
 import (
 	"context"
-	"crypto/rand"
-	"math/big"
 	"sort"
 
 	"github.com/CeresDB/ceresmeta/server/cluster/metadata"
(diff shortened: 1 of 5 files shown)


server/coordinator/shard_picker.go
```diff
@@ -99,3 +100,63 @@ func copyNodeShardMapping(nodeShardsMapping map[string][]storage.ShardNode) map[
 	}
 	return tempNodeShardMapping
 }
+
+// LeastTableShardPicker selects shards based on the number of tables on the current shard,
+// and always selects the shard with the smallest number of current tables.
+type LeastTableShardPicker struct{}
```
ShiKaiWi, line 106: With this picker, shall we remove the random picker?

There is no need to expose the implementation, name it as `leastTableShardPicker`.

server/coordinator/shard_picker.go
```diff
@@ -99,3 +100,63 @@ func copyNodeShardMapping(nodeShardsMapping map[string][]storage.ShardNode) map[
...
+	for i := 0; i < expectShardNum; i++ {
+		if i%len(shardTableNumberMapping) == 0 {
+			sortedMapping = sortByTableNumber(shardTableNumberMapping)
+		}
+
+		selectShard := sortedMapping[i%len(shardTableNumberMapping)]
+		shardTableNumberMapping[selectShard.shardID]++
+
+		result = append(result, shardNodeMapping[selectShard.shardID])
+	}
```
ShiKaiWi, line 139: I guess it can be made in an easier and concise way: Pick the `expectShardNum` shards with the fewest tables, that is to say, there is no need to maintain the shards' table number after one shard selected.

And with this way, the selected `expectShardNum` is also ensured different.

ShiKaiWi (changes requested):

server/coordinator/shard_picker.go
```diff
@@ -19,83 +18,41 @@ type ShardPicker interface {
...
+	result := make([]storage.ShardNode, 0, expectShardNum)
 
-	nodeShardsMapping := make(map[string][]storage.ShardNode, 0)
-	for _, shardNode := range shardNodes {
-		_, exists := nodeShardsMapping[shardNode.NodeName]
-		if !exists {
-			nodeShardsMapping[shardNode.NodeName] = []storage.ShardNode{}
-		}
-		nodeShardsMapping[shardNode.NodeName] = append(nodeShardsMapping[shardNode.NodeName], shardNode)
+	ShardIDs := make([]storage.ShardID, 0, len(snapshot.Topology.ShardViewsMapping))
```
ShiKaiWi, line 41: Rename it as `shardIDs` or `sortedShardsByTableCount`.

server/coordinator/shard_picker.go
```diff
@@ -19,83 +18,41 @@ type ShardPicker interface {
...
+	for _, shardNode := range snapshot.Topology.ClusterView.ShardNodes {
+		shardNodeMapping[shardNode.ID] = shardNode
+	}
+
 	if len(snapshot.Topology.ClusterView.ShardNodes) == 0 {
 		return nil, errors.WithMessage(ErrNodeNumberNotEnough, "no shard is assigned")
 	}
 
-	shardNodes := snapshot.Topology.ClusterView.ShardNodes
+	result := make([]storage.ShardNode, 0, expectShardNum)
```
ShiKaiWi, line 39: Move this line to just above L52.

server/coordinator/shard_picker.go
```diff
@@ -19,83 +18,41 @@ type ShardPicker interface {
...
-
-	nodeShardsMapping := make(map[string][]storage.ShardNode, 0)
-	for _, shardNode := range shardNodes {
-		_, exists := nodeShardsMapping[shardNode.NodeName]
-		if !exists {
-			nodeShardsMapping[shardNode.NodeName] = []storage.ShardNode{}
-		}
-		nodeShardsMapping[shardNode.NodeName] = append(nodeShardsMapping[shardNode.NodeName], shardNode)
+	sortedShardsByTableCount := make([]storage.ShardID, 0, len(snapshot.Topology.ShardViewsMapping))
+	for shardID := range snapshot.Topology.ShardViewsMapping {
```
ShiKaiWi, line 40: Shall we only collect the shards which has been allocated to a node? That is to say, we should loop over the shards of the `shardNodeMapping`. Otherwise, L54 may not find the correct shard node.

server/coordinator/shard_picker.go
```diff
@@ -19,83 +18,41 @@ type ShardPicker interface {
...
-
-		// Remove select shard.
-		nodeShards[selectNodeShardsIndex.Int64()] = nodeShards[len(nodeShards)-1]
-		tempNodeShardMapping[nodeNames[selectNodeIndex.Int64()]] = nodeShards[:len(nodeShards)-1]
-
-		// Remove select node.
-		nodeNames[selectNodeIndex.Int64()] = nodeNames[len(nodeNames)-1]
-		nodeNames = nodeNames[:len(nodeNames)-1]
+		selectShardID := sortedShardsByTableCount[i%len(sortedShardsByTableCount)]
+		result = append(result, shardNodeMapping[selectShardID])
```
ShiKaiWi, line 54:
```suggestion
		shardNode, ok := shardNodeMapping[selectShardID]
		assert.Assert(ok)
		result = append(result, shardNode)
```

ShiKaiWi (approved):

PR #250, opened 2023-10-10
feat: support access etcd with tls

ZuLiangWang:
## Rationale
Currently we support external `ETCD`, However, the configuration of tls is missing, which will result in the inability to connect to the etcd cluster normally.

## Detailed Changes
* Added tls related configuration.

## Test Plan
Test in local environment.


server/server.go
```diff
@@ -4,7 +4,9 @@ package server
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"go.etcd.io/etcd/client/pkg/v3/transport"
```
ShiKaiWi, line 9: Move this to L15?

ShiKaiWi (approved):

PR #251, opened 2023-10-10
fix: TLS config

ZuLiangWang:
## Rationale
Fix wrong tls config.
Modify parameters according to official documentation: https://pkg.go.dev/github.com/coreos/etcd/clientv3#Config

## Detailed Changes
* Fix wrong tls config.

## Test Plan
Verify on etcd server with tls.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -19,10 +19,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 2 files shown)


chunshao90 (approved):
LGTM

PR #254, opened 2023-10-11
feat: support MaxOpsPerTxn for operate etcd

ShiKaiWi:
## Rationale
The external etcd server may have a limit on the number of operations in a transaction. However, CeresMeta don't impose such a limit when operating etcd, and it will lead to error.

## Detailed Changes
Introduce such a limit on the number of operations in a transaction.

## Test Plan
Existing tests.

### DIFF
--- a/server/cluster/metadata/cluster_metadata.go
+++ b/server/cluster/metadata/cluster_metadata.go
@@ -231,7 +231,7 @@ func (c *ClusterMetadata) MigrateTable(ctx context.Context, request MigrateTable
 
 		if !exists {
 			c.logger.Error("the table to be closed does not exist", zap.String("schemaName", request.SchemaName), zap.String("tableName", tableName))
-			return errors.WithMessagef(ErrTableNotFound, "table not exists, shcemaName:%s,tableName:%s", request.SchemaName, tableName)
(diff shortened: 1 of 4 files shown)


server/cluster/metadata/cluster_metadata.go
```diff
@@ -231,7 +231,7 @@ func (c *ClusterMetadata) MigrateTable(ctx context.Context, request MigrateTable
 
 		if !exists {
 			c.logger.Error("the table to be closed does not exist", zap.String("schemaName", request.SchemaName), zap.String("tableName", tableName))
-			return errors.WithMessagef(ErrTableNotFound, "table not exists, shcemaName:%s,tableName:%s", request.SchemaName, tableName)
+			return errors.WithMessagef(ErrTableNotFound, "table not exists, schemaName:%s,tableName:%s", request.SchemaName, tableName)
```
chunshao90, line 234:
```suggestion
			return errors.WithMessagef(ErrTableNotFound, "table not exists, schemaName:%s, tableName:%s", request.SchemaName, tableName)
```

chunshao90 (approved):
LGTM

PR #259, opened 2023-10-16
fix: modify the logic of pick shards

ZuLiangWang:
## Rationale
We found that according to the current way of pick shards, the subtables of the partition table will be concentrated on some nodes, causing hot spots, so some modifications need to be made to avoid this phenomenon.

## Detailed Changes
* Modify the logic of pick shard, when the number of tables is the same, sort according to the size of ShardID.

## Test Plan


chunshao90 (commented):
Add a unit test to check the logic.

jiacai2050 (approved):
LGTM

PR #261, opened 2023-10-19
fix: nil resp if no need to forward

ShiKaiWi:
## Rationale
The response is nil if there is no need to do http fowarding.

## Detailed Changes
Only close the resp body if the http request needs forward.

## Test Plan
Manual test.


ZuLiangWang (approved):
LGTM

PR #265, opened 2023-10-30
refactor!: rename binary to ceresmeta-server

tisonkun:
When install the binary with `go install github.com/CeresDB/ceresmeta/cmd/meta@latest`, the result is `meta` which is too general to conflict with other binaries and easily misleading.

Rename the binary with a clear long name `ceresmeta-server`.

Note that we should adjust ceresdb repo's usage accordingly. If this change gets accepted, I'll handle that.

cc @jiacai2050


scripts/run_integration_test.sh
tisonkun, line 1: I'm going to introduce a BINPATH var for this script so that we don't do all these tricks. It is brought from [tiup's binpath](https://docs.pingcap.com/tidb/stable/tiup-reference#--binpath). But we need to merge this patch first, updating ceresdb code, and update this script finally.

ZuLiangWang (approved):
LGTM

scripts/run_integration_test.sh
ZuLiangWang: Thanks for your contribution. I will resolve this conversation first to merge this PR.

ZuLiangWang:
CeresDB also needs to compile CeresMeta to complete the cluster integration test. Maybe we need to adapt this file: https://github.com/CeresDB/ceresdb/blob/main/integration_tests/build_meta.sh

tisonkun:
Yes. Let me preapre a patch for ceresdb today.

tisonkun:
FYI - https://github.com/CeresDB/ceresdb/pull/1287/files

PR #269, opened 2023-11-02
ci: check license header tool

tisonkun:
Use hawkeye to check license headers. Apply with the standard Apache License 2.0 header.

You can review in commits and skip the globally apply changes one.


tisonkun:
cc @ShiKaiWi @ZuLiangWang

tanruixiang (approved):
LGTM. It looks like we have a problem with the inception year in the [Ceresdb](https://github.com/CeresDB/ceresdb/blob/main/licenserc.toml) license header and should change it to 2022. 🤔

Makefile
```diff
@@ -1,7 +1,3 @@
-# Refers to https://github.com/tikv/pd/blob/master/Makefile
-
-default: build
-
```
tanruixiang, line 4: What I'm not sure about is, do we need to keep the link?

tisonkun: Maybe. But I don't find any similarities this Makefile with pd's.

tanruixiang: @chunshao90 Do you remember why we added this link before? 🤣

ShiKaiWi: I guess there are some similarities between them: https://github.com/tikv/pd/blob/master/Makefile#L143

tisonkun: They are quite different and if it's not clearly driven from another file, we should try to reduce the IP burden.

ShiKaiWi (approved):

Makefile
tanruixiang: > They are quite different and if it's not clearly derived from another file, we should try to reduce the IP burden.

Makes sense to me.

PR #271, opened 2023-11-10
chore: Ratelimiter is enabled by default

chunshao90:
## Rationale
Ratelimiter is enabled by default to prevent ceresmeta from being hung.

## Detailed Changes
Ratelimiter is enabled by default.

## Test Plan
CI.

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -38,9 +38,9 @@ const (
 	defaultEtcdKeyPath          = ""
 	defaultEtcdCertPath         = ""
 
+	defaultEnableLimiter          bool  = true
 	defaultInitialLimiterCapacity int   = 100 * 1000
 	defaultInitialLimiterRate     int   = 10 * 1000
-	defaultEnableLimiter          bool  = false
 	defaultEtcdStartTimeoutMs     int64 = 60 * 1000
 	defaultCallTimeoutMs                = 5 * 1000
 	defaultEtcdMaxTxnOps                = 128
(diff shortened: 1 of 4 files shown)


server/limiter/limiter.go
```diff
@@ -24,26 +24,26 @@ import (
 )
 
 type FlowLimiter struct {
-	l *rate.Limiter
+	// enable is used to control the switch of the limiter.
+	enable bool
+	l      *rate.Limiter
```
Rachelint, line 29: How about naming it more clearly...

server/config/config.go
```diff
@@ -38,9 +38,9 @@ const (
 	defaultEtcdKeyPath          = ""
 	defaultEtcdCertPath         = ""
 
+	defaultEnableLimiter          bool  = true
 	defaultInitialLimiterCapacity int   = 100 * 1000
```
ZuLiangWang, line 42: If the `RateLimiter` is turned on by default, we need to carefully consider the setting of the default value for `RateLimiter`. An excessively large threshold will cause the limiter to be effectively invalid.

chunshao90: By default, the rate limit threshold is set to 10000, which should be quite large. However, you can adjust the rate limit according to your specific needs and business requirements.

ZuLiangWang (approved):
LGTM

server/limiter/limiter.go
chunshao90: Refactor in another PR.

PR #273, opened 2023-11-13
refactor: rename deployMode to enableSchedule

chunshao90:
## Rationale
`DeployMode` is not a good name, it is not easily understandable. I suggest changing it to `enableSchedule`.

## Detailed Changes
Change `DeployMode` to `EnableSchedule`.

## Test Plan
CI.


ZuLiangWang (approved):
LGTM

PR #276, opened 2023-11-14
fix: procedure id path

chunshao90:
## Rationale
The procedure ID path does not include the Ceresmeta prefix.

## Detailed Changes
Add the Ceresmeta prefix to the procedure ID path.

## Test Plan
CI.

### DIFF
--- a/server/cluster/cluster.go
+++ b/server/cluster/cluster.go
@@ -53,7 +53,7 @@ func NewCluster(logger *zap.Logger, metadata *metadata.ClusterMetadata, client *
 		return nil, errors.WithMessage(err, "create procedure manager")
 	}
 	dispatch := eventdispatch.NewDispatchImpl()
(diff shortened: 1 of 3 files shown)


server/cluster/cluster.go
```diff
@@ -108,3 +109,7 @@ func (c *Cluster) GetShards() []storage.ShardID {
 func (c *Cluster) GetShardNodes() metadata.GetShardNodesResult {
 	return c.metadata.GetShardNodes()
 }
+
+func getProcedureIDPath(rootPath string) string {
```
ShiKaiWi, line 113: Maybe we should use `/{rootPath}/{clusterName}/{prefixKey}` to make it separate from different clusters.

ShiKaiWi, line 113: `make` or `format` seems better than `get`:
```suggestion
func makeProcedureIDPath(rootPath string) string {
```

chunshao90: Currently, the procedure design is shared for Ceresmeta and does not involve cluster-level differentiation. However, if you want to add a cluster prefix, the procedure storage would also need to include the prefix.

ShiKaiWi (approved):

PR #279, opened 2023-11-17
refactor: procedure storage

ZuLiangWang:
## Rationale
The current design of the procedure storage module is unreasonable. The data will be stored permanently and will not be cleared. This will cause the storage space of `ETCD` to be filled up quickly. We need to redesign the interface of procedure storage.

## Detailed Changes
* Add `ttl` config for `CreateOrUpdate`.
* Add `Delete` interface, it used to delete history procedure data and current procedure data.

## Test Plan

### DIFF
--- a/server/coordinator/procedure/procedure.go
+++ b/server/coordinator/procedure/procedure.go
@@ -32,11 +32,11 @@ const (
 	StateCancelled = "cancelled"
 )
 
-type Typ uint
+type Kind uint
 
 const (
 	// Cluster Operation
-	Create Typ = iota
+	Create Kind = iota
 	Delete
 	TransferLeader
 	Migrate
@@ -65,8 +65,8 @@ type Procedure interface {
 	// ID of the procedure.
 	ID() uint64
 
-	// Typ of the procedure.
-	Typ() Typ
+	// Kind of the procedure.
+	Kind() Kind
 
 	// Start the procedure.
 	Start(ctx context.Context) error
@@ -87,7 +87,7 @@ type Procedure interface {
 // Info is used to provide immutable description procedure information.
 type Info struct {
 	ID    uint64
-	Typ   Typ
+	Typ   Kind
 	State State
 }
 
--- a/server/coordinator/procedure/procedure.go
+++ b/server/coordinator/procedure/procedure.go
@@ -87,7 +87,7 @@ type Procedure interface {
 // Info is used to provide immutable description procedure information.
 type Info struct {
 	ID    uint64
-	Typ   Kind
+	Kind  Kind
 	State State
 }
 
--- a/server/coordinator/procedure/storage.go
+++ b/server/coordinator/procedure/storage.go
@@ -21,7 +21,8 @@ import (
 )
 
 type Write interface {
-	CreateOrUpdate(ctx context.Context, meta Meta, ttlSec int64) error
+	CreateOrUpdate(ctx context.Context, meta Meta) error
+	CreateOrUpdateWithTTL(ctx context.Context, meta Meta, ttlSec int64) error
 }
 
 type Meta struct {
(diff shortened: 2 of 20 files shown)


server/coordinator/procedure/storage.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 type Write interface {
-	CreateOrUpdate(ctx context.Context, meta Meta) error
+	CreateOrUpdate(ctx context.Context, meta Meta, ttl int64) error
```
ShiKaiWi, line 24: `ttl` is ambiguous for callers, how about put it as:
```suggestion
	CreateOrUpdate(ctx context.Context, meta Meta, ttlSec int64) error
```

server/coordinator/procedure/storage_impl.go
```diff
@@ -49,31 +53,66 @@ func NewEtcdStorageImpl(client *clientv3.Client, rootPath string, clusterID uint
...
-	keyPath := e.generaNormalKeyPath(id)
+func (e EtcdStorageImpl) MarkDeleted(ctx context.Context, procedureType Typ, id uint64) error {
+	keyPath := e.generaNormalKeyPath(procedureType, id)
 	meta, err := etcdutil.Get(ctx, e.client, keyPath)
 	if err != nil {
 		return errors.WithMessage(err, "get meta failed")
 	}
 
-	deletedKeyPath := e.generaDeletedKeyPath(id)
+	deletedKeyPath := e.generaDeletedKeyPath(procedureType, id)
```
ShiKaiWi, line 115: Could you give an explanation about why the type of procedure is part of the key? It looks very weird.

ZuLiangWang: Considering query performance, currently we can only query all procedures through `List`. Using `Type` as part of the key can effectively reduce the amount of data queried in `list`.

server/coordinator/procedure/storage.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 type Write interface {
-	CreateOrUpdate(ctx context.Context, meta Meta) error
+	CreateOrUpdate(ctx context.Context, meta Meta, ttlSec int64) error
```
ShiKaiWi, line 24: How about adding a new method called `CreateOrUpdateWithTTL` and keep the `CreateOrUpdate` same as before? If so, there is no need to change the irrelevant codes and there is no need to use zero ttlSec as a sentinel for no ttl.

server/coordinator/procedure/storage_impl.go
```diff
@@ -49,31 +53,66 @@ func NewEtcdStorageImpl(client *clientv3.Client, rootPath string, clusterID uint
...
 
 	if _, err = e.client.Do(ctx, opPut); err != nil {
 		return errors.WithMessage(err, "etcd put data failed")
 	}
+
+	return nil
+}
+
+// Delete will delete the specified procedure,
+// and try to delete its corresponding history procedure if it exists.
```
ShiKaiWi, line 86: The style of the comment seems weird, how about letting the formatter break the original one line into two lines?

server/coordinator/procedure/procedure.go
```diff
@@ -87,7 +87,7 @@ type Procedure interface {
 // Info is used to provide immutable description procedure information.
 type Info struct {
 	ID    uint64
-	Typ   Typ
+	Typ   Kind
```
ShiKaiWi, line 90: Kind？

server/coordinator/procedure/storage.go
```diff
@@ -22,17 +22,19 @@ import (
 
 type Write interface {
 	CreateOrUpdate(ctx context.Context, meta Meta) error
+	CreateOrUpdateWithTTL(ctx context.Context, meta Meta, ttlSec int64) error
 }
 
 type Meta struct {
 	ID      uint64
-	Typ     Typ
+	Typ     Kind
```
ShiKaiWi, line 30: Kind？

ShiKaiWi (approved):

PR #286, opened 2023-12-05
refactor: make create table idempotent

ZuLiangWang:
## Rationale
Store the intermediate data for table creation through procedures to support idempotent table creation.

## Detailed Changes
* Store the intermediate data for table creation through procedure.
* Supports retrying from the upper layer to complete failed table creation procedures.

## Test Plan

### DIFF
--- a/server/cluster/metadata/topology_manager.go
+++ b/server/cluster/metadata/topology_manager.go
@@ -44,6 +44,8 @@ type TopologyManager interface {
 	AddTable(ctx context.Context, shardID storage.ShardID, latestVersion uint64, tables []storage.Table) error
 	// RemoveTable remove table on target shards from cluster topology.
 	RemoveTable(ctx context.Context, shardID storage.ShardID, latestVersion uint64, tableIDs []storage.TableID) error
+	// GetTableShard get the shardID of the shard where the table is located.
+	GetTableShard(ctx context.Context, table storage.Table) (storage.ShardID, bool)
 	// AssignTable persistent table shard mapping, it is used to store assign results and make the table creation idempotent.
 	AssignTable(ctx context.Context, schemaID storage.SchemaID, tableName string, shardID storage.ShardID) error
 	// GetAssignTableResult get table assign result.
@@ -158,6 +160,7 @@ func NewTopologyManagerImpl(logger *zap.Logger, storage storage.Storage, cluster
 		nodeShardsMapping:  nil,
 		shardTablesMapping: nil,
 		tableShardMapping:  nil,
+		tableAssignMapping: nil,
 		nodes:              nil,
 	}
 }
@@ -307,9 +310,21 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
 	return nil
 }
 
-func (m *TopologyManagerImpl) AssignTable(ctx context.Context, schemaID storage.SchemaID, tableName string, shardID storage.ShardID) error {
+func (m *TopologyManagerImpl) GetTableShard(_ context.Context, table storage.Table) (storage.ShardID, bool) {
 	m.lock.RLock()
 	defer m.lock.RUnlock()
+
+	shardIDs, exists := m.tableShardMapping[table.ID]
+	if exists {
+		return shardIDs[0], true
+	}
+
+	return 0, false
+}
+
+func (m *TopologyManagerImpl) AssignTable(ctx context.Context, schemaID storage.SchemaID, tableName string, shardID storage.ShardID) error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
 
 	if err := m.storage.AssignTable(ctx, storage.AssignTableRequest{
 		ClusterID: m.clusterID,
@@ -321,6 +336,10 @@ func (m *TopologyManagerImpl) AssignTable(ctx context.Context, schemaID storage.
 	}
 
 	// Update cache im memory.
+	if _, exists := m.tableAssignMapping[schemaID]; !exists {
+		m.tableAssignMapping[schemaID] = make(map[string]storage.ShardID, 0)
+	}
+
 	m.tableAssignMapping[schemaID][tableName] = shardID
 
 	return nil
@@ -332,8 +351,8 @@ func (m *TopologyManagerImpl) GetAssignTableResult(_ context.Context, schemaID s
 }
 
 func (m *TopologyManagerImpl) DeleteAssignTable(ctx context.Context, schemaID storage.SchemaID, tableName string) error {
-	m.lock.RLock()
-	defer m.lock.RUnlock()
+	m.lock.Lock()
+	defer m.lock.Unlock()
 
 	if err := m.storage.DeleteAssignTable(ctx, storage.DeleteAssignTableRequest{
 		ClusterID: m.clusterID,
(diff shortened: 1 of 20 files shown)


server/cluster/metadata/topology_manager.go
```diff
@@ -291,6 +307,64 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
 	return nil
 }
 
+func (m *TopologyManagerImpl) GetTableShard(_ context.Context, table storage.Table) (storage.ShardID, bool) {
```
ShiKaiWi, line 310: If the fetched result is shard id, how about renaming it as:
```suggestion
func (m *TopologyManagerImpl) GetTableShardID(_ context.Context, table storage.Table) (storage.ShardID, bool) {
```

server/coordinator/procedure/ddl/createtable/create_table.go
```diff
@@ -34,37 +34,113 @@ import (
...
+	schemaName := params.SourceReq.GetSchemaName()
+	tableName := params.SourceReq.GetName()
+
+	targetShardID, exists, err := params.ClusterMetadata.GetAssignTable(req.ctx, schemaName, tableName)
+	if err != nil {
+		procedure.CancelEventWithLog(event, err, "get table assign", zap.String("schemaName", schemaName), zap.String("tableName", tableName))
+		return
+	}
+	if exists {
+		if targetShardID != params.ShardID {
```
ShiKaiWi, line 111: How about picking the shard here? After that, such duplicate codes with the ones in `factory.go` can be avoided, and there is no such `ShardNotMatch` error.

ShiKaiWi:
@ZuLiangWang After a bit of reviewing, I find the table id is not persisted in the `TableAssign` procedure, and table creation may fail if a new table id is allocated in the second retry. And is it true?

ShiKaiWi:
> @ZuLiangWang After a bit of reviewing, I find the table id is not persisted in the `TableAssign` procedure, and table creation may fail if a new table id is allocated in the second retry. And is it true?

After digging into the code not included by this PR, I find the table id will be fetched from the table meta data, so the second creation will use the same table id as before.

server/cluster/metadata/topology_manager.go
```diff
@@ -291,6 +307,64 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
...
+
+	shardIDs, exists := m.tableShardMapping[table.ID]
+	if exists {
+		return shardIDs[0], true
+	}
+
+	return 0, false
+}
+
+func (m *TopologyManagerImpl) AssignTable(ctx context.Context, schemaID storage.SchemaID, tableName string, shardID storage.ShardID) error {
```
ShiKaiWi, line 322: This method actually assigns the table to a picked shard, so how about making the name more explicit:
```suggestion
func (m *TopologyManagerImpl) AssignTableToShard(ctx context.Context, schemaID storage.SchemaID, tableName string, shardID storage.ShardID) error {
```

server/cluster/metadata/topology_manager.go
```diff
@@ -291,6 +307,64 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
...
+	if _, exists := m.tableAssignMapping[schemaID]; !exists {
+		m.tableAssignMapping[schemaID] = make(map[string]storage.ShardID, 0)
+	}
+
+	m.tableAssignMapping[schemaID][tableName] = shardID
+
+	return nil
+}
+
+func (m *TopologyManagerImpl) GetAssignTableResult(_ context.Context, schemaID storage.SchemaID, tableName string) (storage.ShardID, bool) {
```
ShiKaiWi, line 345:
```suggestion
func (m *TopologyManagerImpl) GetTableAssignedShard(_ context.Context, schemaID storage.SchemaID, tableName string) (storage.ShardID, bool) {
```

server/cluster/metadata/topology_manager.go
```diff
@@ -291,6 +307,64 @@ func (m *TopologyManagerImpl) RemoveTable(ctx context.Context, shardID storage.S
...
+	assignResult, exists := m.tableAssignMapping[schemaID][tableName]
+	return assignResult, exists
+}
+
+func (m *TopologyManagerImpl) DeleteAssignTable(ctx context.Context, schemaID storage.SchemaID, tableName string) error {
```
ShiKaiWi, line 350:
```suggestion
func (m *TopologyManagerImpl) DeleteTableAssignedShard(ctx context.Context, schemaID storage.SchemaID, tableName string) error {
```

server/coordinator/factory.go
```diff
@@ -126,22 +126,32 @@ func (f *Factory) makeCreateTableProcedure(ctx context.Context, request CreateTa
...
+		shards, err := f.shardPicker.PickShards(ctx, snapshot, 1)
+		if err != nil {
+			f.logger.Error("pick table shard", zap.Error(err))
+			return nil, errors.WithMessage(err, "pick table shard")
+		}
+		if len(shards) != 1 {
+			f.logger.Error("pick table shards length not equal 1", zap.Int("shards", len(shards)))
+			return nil, errors.WithMessagef(procedure.ErrPickShard, "pick table shard, shards length:%d", len(shards))
+		}
+		targetShardID = shards[0].ID
```
ShiKaiWi, line 146: I guess we can encapsulate the table assignment to the `ShardPicker`. Just provide a new `ShardPicker` wrapper around the original one, and do table assignment query/creation in the wrapper.

server/coordinator/persist_shard_picker.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func NewPersistShardPicker(cluster *metadata.ClusterMetadata, internal ShardPicker) *PersistShardPicker {
+	return &PersistShardPicker{cluster: cluster, internal: internal}
+}
+
+func (p *PersistShardPicker) PickShards(ctx context.Context, snapshot metadata.Snapshot, schemaName string, tableNames []string) (map[string]storage.ShardNode, error) {
+	result := map[string]storage.ShardNode{}
+
+	shardNodeMap := map[storage.ShardID]storage.ShardNode{}
```
ShiKaiWi, line 41: Maybe we should pre-allocate the memory.

server/coordinator/persist_shard_picker.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if len(result) == len(tableNames) {
+		return result, nil
+	}
+
+	var tablesNeedToAssignShard []string
+	if len(missingTables) > 0 {
+		tablesNeedToAssignShard = missingTables
+	} else {
+		tablesNeedToAssignShard = tableNames
```
ShiKaiWi, line 68: This branch seems unreachable?

server/coordinator/persist_shard_picker.go
```diff
@@ -0,0 +1,86 @@
...
+			return map[string]storage.ShardNode{}, err
+		}
+		if exists {
+			result[tableNames[i]] = shardNodeMap[shardID]
+		} else {
+			missingTables = append(missingTables, tableNames[i])
+		}
+	}
+
+	if len(result) == len(tableNames) {
```
ShiKaiWi, line 60: Maybe we should check whether `missingTables` is empty? If the any table name is duplicate in the `tableNames`, this check may won't work.

ShiKaiWi (approved):

PR #292, opened 2023-12-22
chore: add asf yaml

chunshao90:
## Rationale
Add `.asf.yaml`.

## Test Plan
No need.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -6,7 +6,7 @@
 # "License"); you may not use this file except in compliance
 # with the License.  You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


caicancai:
I think we should add a cli check for icense. I don’t know if horaedb currently has this measure.

chunshao90:
> I think we should add a cli check for icense. I don’t know if horaedb currently has this measure.

license checker？

caicancai:
> > I think we should add a cli check for icense. I don’t know if horaedb currently has this measure.
> 
> license checker？

yes

tanruixiang (approved):
LGTM

chunshao90:
> I think we should add a cli check for icense. I don’t know if horaedb currently has this measure.

License checker already exists in horaemeta.

PR #295, opened 2023-12-28
feat: add drop shard node api

ZuLiangWang:
## Rationale
In some extreme cases (for this situation, please refer to https://github.com/apache/incubator-horaedb-meta/issues/296), the shard node cannot be cleaned normally, and we need to add an operation and maintenance interface to complete manual cleaning.

## Detailed Changes
* Add drop shard node api.

## Test Plan
Pass CI.


jiacai2050 (approved):
LGTM

PR #298, opened 2024-01-04
chore!: replace ceresdb with test

jiacai2050:
## Rationale
See https://github.com/apache/incubator-horaedb/issues/1319

## Detailed Changes
This is a breaking change. users need to add following config to keep back-compatitble
```
storage-root-path = "/ceresdb"
```
Or set via env vars 
```
export STORAGE_ROOT_PATH="/ceresdb"
```
## Test Plan
No need


ZuLiangWang (approved):
LGTM

PR #300, opened 2024-01-10
refactor: remove useless procedure persist

ZuLiangWang:
## Rationale
Currently, the `TransferLeader` procedure will be persisted every time it is executed, but they actually have no effect. When the cluster runs for a long time, some garbage data will be generated to occupy ETCD's storage space.

## Detailed Changes
* Remove useless procedure persist in `transfer_leader`.

## Test Plan
Pass CI.


jiacai2050 (approved):
LGTM

